PR #19, opened 2017-12-21
Removal of empty declarations

jolkdarr:
Useless empty declarations should be removed.


mvo5:
The fix is appreciated. I'm curious, are you working/plan to work on more synaptic fixes?

PR #41, opened 2019-06-17
Port to apt 1.9.0

julian-klode:


julian-klode:
This also accidentally fixes some bugs:

- Archive cleaning works now (was a no-op before, as Erase did nothing)
- The code that used GetCandidateVer now uses GetCandidateVersion, so respects policy

julian-klode:
Force-pushed as a rebase went a bit wrong, and a part was in a wrong commit.

Tested running with 1.9.0, and compile-tested with 1.8.2.

mvo5:
Thanks for this port! I plan to merge this and release synaptic 0.85 with. The only reason I have not done it yet is that it will probably break people building synaptic with apt-rpm. I'm not sure though if how much of an issue this is, I will try to find out.

PR #53, opened 2020-04-06
Fix one-line-tall diff viewer

andrewshadura:
Hi,

This fixes the Debian bug #[944219](https://bugs.debian.org/944219)


mvo5:
@andrewshadura nice, thank you!

PR #54, opened 2020-05-06
Avoid FTBFS in Ubuntu

ginggs:
https://bugs.debian.org/958960

This patch no no longer applies, and no longer seems necessary, since #45 was merged.


ginggs:
We can probably remove all of the dpatch-related stuff at this point

mvo5:
Thanks for this PR. This got fixed in synaptic 0.90.1

PR #59, opened 2020-11-11
Avoid dpatch, update pre-build.sh for configure.ac

julian-klode:
Updates to avoid the dpatching, now that CTTE said that dpkg distro patches are not allowed, same rationale applies here. Instead use dpkg-vendor in the configure.ac to determine if we are deriving from Ubuntu and `#ifdef` accordingly.

Update pre-build script for configure.ac instead of configure.in, and make it fail if it is modifying the file so you get to update your git repo.

Tested on Ubuntu hirsute.


julian-klode:
Oh yes, this also adds ESM as supported, as it should be.

mvo5 (approved):
Nice, thank you

PR #62, opened 2020-12-09
Drop gnome-common from autogen, also remove macros/ directory

BryanQuigley:
Follow https://wiki.gnome.org/Projects/GnomeCommon/Migration

Also drop macros/ directory which may not be used at all for a while?

From what I can tell nothing appears to break from these changes. It is certainly possible I missed something.


mvo5:
Thanks for this, really apprectated! This looks fine.

PR #63, opened 2020-12-16
Updated Readme and add initial Contributing.md

BryanQuigley:
- Simplified history
- Removed mentioning when features were added (that have been here
for a long time)
- use current package description as README intro
- other small tweaks to better align to markdown


mvo5 (approved):
Thank you, much nicer this way!

PR #64, opened 2020-12-30
Update all the home page mentions that appear current

### LINKED ISSUE #61: Link to this github as homepage instead of www.nongnu.org/synaptic/
I'd also be happy to create a new homepage using github pages if interested.  Either way, it should link to something that is actively maintained.

BryanQuigley:
This resolves #61


mvo5 (approved):
Thank you, this looks great

PR #65, opened 2020-12-30
Drop WINGS gui

BryanQuigley:
From what I can see WINGS hasn't been built in 15 years, perhaps it can be dropped?


Amr-Ibra:
Gentle ping @mvo5

mvo5:
Hey, sorry that I took so long for this one - this old code has a bit of sentimental value but of course it's a bit silly to hold on to it. It was indeed not shipped in the debs for the last 15y or so. In theory someone might use it independently that I'm not aware of but I doubt it. We could always revert if that is the case I guess. So yeah, let's do it.

Thanks for the PR!

mvo5 (approved):
Thank you

PR #66, opened 2020-12-30
Drop gtk2

BryanQuigley:
With GTK4s release (https://blog.gtk.org/2020/12/16/gtk-4-0/), GTK2 is soon to be EOL.  Let's drop compatibly with it.

We could also likely drop support for the first few versions of gtk3 (3.2/3.4), but that;s not included.


gtk/rgdebinstallprogress.cc
```diff
@@ -293,14 +292,6 @@ void RGDebInstallProgress::startUpdate()
    // check if we run embedded
    int id = _config->FindI("Volatile::PlugProgressInto", -1);
    if (id > 0) {
-#if !GTK_CHECK_VERSION(3,0,0)
-      GtkWidget *vbox = GTK_WIDGET(gtk_builder_get_object(_builder, "vbox_rgdebinstall_progress"));
-      _sock =  gtk_plug_new(id);
-      gtk_widget_reparent(vbox, _sock);
-      gtk_widget_show(_sock);
-#else
```
mvo5, line 301: We need to keep the "else" part here to indicate that this is not supported. I guess we could also just drop this feature entirely, i.e. remove the "int id = " and the if part and just keep "show();". But best to do this as a separate PR to keep the changes that are very mechanical (like doing #if/#then) separate from the ones that actually change behavior. Does that make sense?

gtk/rgfetchprogress.cc
```diff
@@ -133,18 +133,6 @@ RGFetchProgress::RGFetchProgress(RGWindow *win)
...
-   if (id > 0) {
-#if !GTK_CHECK_VERSION(3,0,0)
-      gtk_widget_hide(GTK_WIDGET(gtk_builder_get_object(_builder, "window_fetch")));
-      GtkWidget *vbox = GTK_WIDGET(gtk_builder_get_object(_builder, "vbox_fetch"));
-      _sock =  gtk_plug_new(id);
-      gtk_widget_reparent(vbox, _sock);
-      gtk_widget_show_all(_sock);
-      _win = _sock;
-#else
-      g_error("gtk_plugin_new not supported with gtk3");
```
mvo5, line 145: Similar as above, I think we should either keep the "g_error" here or we drop the "int id = " too because it is only needed for this. Ideally we would also document that "Volatile::PlugProgressInto" is no longer working but given that it was only working with gtk2 anyway and that gets dropped I doubt it's an issue :)

gtk/rgterminstallprogress.cc
```diff
@@ -129,17 +129,6 @@ void RGTermInstallProgress::startUpdate()
...
-#if !GTK_CHECK_VERSION(3,0,0)
-      gtk_widget_hide(GTK_WIDGET(gtk_builder_get_object
-                                 (_builder, "window_zvtinstallprogress")));
-      GtkWidget *vbox = GTK_WIDGET(gtk_builder_get_object
-                                   (_builder, "vbox_terminstall_progress"));
-      _sock =  gtk_plug_new(id);
-      gtk_widget_reparent(vbox, _sock);
-      gtk_widget_show(_sock);
-#else
-      g_error("gtk_plugin_new not supported with gtk3");
```
mvo5, line 141: Similar as above, let's keep the g_error for now and remove the entire "Volatile::PlugProgressInto" in a followup PR.

mvo5 (commented):
Thanks Bryan for doing this cleanup! I have some small inline comments but the direction looks good. We should probably also document in NEWS somemthing like "dropping gtk2".

BryanQuigley:
Added g_error back.  I think you are right it makes sense to drop them in a separate PR.  Is NEWS still relevant - it hasn't been updated in a while - happy to update it if wanted.

Does GTK3.4 make sense as a new minimum?

gtk/rgrepositorywin.h
```diff
@@ -31,12 +31,6 @@
 #include "rsources.h"
 #include "rggtkbuilderwindow.h"
 
-#if !GTK_CHECK_VERSION(3,0,0)
- #include "rgvendorswindow.h"
-#else
- #warning "no rgvendorswindow in gtk3"
```
mvo5, line 37: We probably want to keep this too or even make it an error - the whole rgrepositorywin.h is not needed for ubuntu/debian iirc, it's a feature of apt-rpm which never got added to the deb based apt. So we could in a followup remove the entire thing or at least mark it very clearly as deprecated. I will merge your PR anyway because the rest looks very nice :)

mvo5:
@BryanQuigley Thanks again for this PR! One small comment inline and I think gtk3.4 sounds fine as the minimal version. NEWS was not touched for a while but might be a good time to resurrect, I mostly did not use it because I did not have time to properly maintain this :/

PR #67, opened 2020-12-30
This just brings the debian patch to git from

BryanQuigley:
This just pushes your patch from Debian to here.  Guessing it's in unpushed local branch?  Feel free to close this and push that.

  [ Michael Vogt ]
  * gtk: fix invalid assignment from SizeToStr to intermediate value
    (Closes: 974790, 974725)


mvo5 (approved):
Nice, thanks for this :)

PR #68, opened 2021-01-04
drop rgvendorswindow*

BryanQuigley:
Unused per last commit and the great majority of was commented out anyway.

From previous comment this has to do with apt-rpm which appears defunct.


mvo5:
This looks fine, I would like to test build it first - which reminds me - maybe we could have a github action for synaptic that just builds synaptic as a first test? or maybe even a test that builds the deb package to ensure we also do not break the packaging.  Do you think that would be something you could look at?

Obviously unit/integration tests beyond this would great too but synaptic is from a time when this was rare so it's not easy to add.

mvo5:
And THANK YOU :)

PR #69, opened 2021-01-04
Remove plug-progress-into option

BryanQuigley:
Only did something in gtk2 and afaict from https://codesearch.debian.net/
the option doesn't seem to be used by anyone else.


mvo5 (approved):
Thank you! Would be nice to mention (e.g. in NEWS) that support was dropped. But not a big deal as it's almost certain that noone was using this for years.

PR #71, opened 2021-01-06
Github actions: add a basic make and debuild

BryanQuigley:


.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,32 @@
...
+  build:
+
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: update
+      run: sudo apt update
+    - name: install packages
+      run: sudo apt-get install --no-install-recommends -y debhelper-compat devscripts libapt-pkg-dev gettext libgtk-3-dev libvte-2.91-dev intltool xmlto sharutils libept-dev
```
mvo5, line 19: Does "sudo apt build-dep -y ./"  also work here? This would have the nice benefit that it also tests the build-dependencies in debian/control.

BryanQuigley:
You can see a build of it here: https://github.com/BryanQuigley/synaptic/pull/2

You may have to enable it for this project in settings -> actions.

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,32 @@
...
+      # libsm-dev lsb-release (already installed on image)
+    - name: autogen.sh
+      run: ./autogen.sh
+    - name: configure
+      run: ./configure
+    - name: make
+      run: make
+    - name: debuild
+      run: debuild -B -us -uc
+#    - name: make check
```
mvo5, line 29: I guess we could try installing it too (and removing it) to check for postinst/postrm issues.

mvo5 (approved):
This is super nice, thank you. Some ideas inline but let's merge it and we can always do more in a followup :)

mvo5:
And the GH action is running in https://github.com/mvo5/synaptic/runs/1658147623?check_suite_focus=true - amazing :) And any new PR should run actions now too.

.github/workflows/c-cpp.yml
BryanQuigley: Yes it does!  I didn't know you could do that :)

PR #72, opened 2021-01-06
add install/removing the deb

BryanQuigley:
also test the binaries in the simplest way possible (does -h work)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -13,11 +13,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: update
-      run: sudo apt update
-    - name: install packages
-      run: sudo apt-get install --no-install-recommends -y debhelper-compat devscripts libapt-pkg-dev gettext libgtk-3-dev libvte-2.91-dev intltool xmlto sharutils libept-dev
-      # libsm-dev lsb-release (already installed on image)
+    - name: update and install packages
+      run: |
+        sudo apt update
+        sudo apt build-dep -y ./
+        #Devscripts for bringing in debuild
+        sudo apt install devscripts -y
+
     - name: autogen.sh
       run: ./autogen.sh
     - name: configure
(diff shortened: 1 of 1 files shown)


.github/workflows/c-cpp.yml
```diff
@@ -13,11 +13,12 @@ jobs:
...
-    - name: update
-      run: sudo apt update
-    - name: install packages
-      run: sudo apt-get install --no-install-recommends -y debhelper-compat devscripts libapt-pkg-dev gettext libgtk-3-dev libvte-2.91-dev intltool xmlto sharutils libept-dev
-      # libsm-dev lsb-release (already installed on image)
+    - name: update and install packages
+      run: |
+        sudo apt update
+        sudo apt build-dep -y ./
+        sudo apt install devscripts -y
```
mvo5, line 20: I assume this is needed for having the `debuild` command? If so, maybe you could add a small comment?

.github/workflows/c-cpp.yml
```diff
@@ -26,7 +27,8 @@ jobs:
       run: make
     - name: debuild
       run: debuild -B -us -uc
-#    - name: make check
-#      run: make check
-#    - name: make distcheck
-#      run: make distcheck
+    - name: install deb
+      run: sudo apt install ../synaptic_*.deb
```
mvo5, line 31: Probably best to add a `-y` here too so that this won't get stuck if it needs extra dependencies. Or is that not a concern (maybe I'm missing something :) ?

.github/workflows/c-cpp.yml
```diff
@@ -26,7 +27,8 @@ jobs:
...
+    - name: remove deb
+      run: sudo apt remove synaptic
+# TODO ? add make check or distcheck?
```
mvo5, line 34: Thanks for adding this TODO

mvo5 (commented):
Thanks, this is really nice. Two small questions/suggestions inline, then it should be good to go.

.github/workflows/c-cpp.yml
BryanQuigley: Apparently it installs dependencies without the -y..  which is somewhat surprising, but it works..

.github/workflows/c-cpp.yml
BryanQuigley: done :)

.github/workflows/c-cpp.yml
mvo5: Heh, interesting. I suspect a bug in apt here but it's fine. If the behavior in apt changes we can update or workflow trivially :)

PR #73, opened 2021-01-09
Move to newer vte bits and fully disable some ifdef 0 code.

BryanQuigley:
Small fixes to remove some of the warnings when building.

### DIFF
--- a/gtk/rgdebinstallprogress.cc
+++ b/gtk/rgdebinstallprogress.cc
@@ -510,7 +510,7 @@ void RGDebInstallProgress::terminalAction(GtkWidget *terminal,
         TermAction action) {
    switch(action) {
       case EDIT_COPY:
-          vte_terminal_copy_clipboard(VTE_TERMINAL(terminal));
+          vte_terminal_copy_clipboard_format(VTE_TERMINAL(terminal), VTE_FORMAT_TEXT);
           break;
       case EDIT_SELECT_ALL:
           vte_terminal_select_all(VTE_TERMINAL(terminal));
--- a/common/rpackage.cc
+++ b/common/rpackage.cc
@@ -1109,9 +1109,9 @@ void RPackage::setPinned(bool flag)
 
 
 // FIXME: this function is broken right now (and it never really wasn't :/
+#if 0
 bool RPackage::isShallowDependency(RPackage *pkg)
 {
-#if 0
    pkgCache::DepIterator rdepI;
 
    // check whether someone else depends on a virtual pkg of this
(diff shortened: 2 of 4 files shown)


BryanQuigley:
Getting fully warning free would mean moving to gtk 3.22 ... 
that would mean no more building on Ubuntu 16.04

gtk/rgdebinstallprogress.cc
```diff
@@ -510,7 +510,7 @@ void RGDebInstallProgress::terminalAction(GtkWidget *terminal,
         TermAction action) {
    switch(action) {
       case EDIT_COPY:
-          vte_terminal_copy_clipboard(VTE_TERMINAL(terminal));
+          vte_terminal_copy_clipboard_format(VTE_TERMINAL(terminal), VTE_FORMAT_TEXT);
```
mvo5, line 513: I just had a quick look and it seems vte_terminal_copy_clipboard_format() was added in 0.5 and it seems in xenial we have 0.42.5. Is that what you mean that for getting fully warning free we would stop building on xenial?

mvo5 (commented):
Thanks! One question about compatibility inline. Note that dropping xenial is not a blocker. I was extremely conservative in the past about keeping compatibility with older versions for distros. But that is not set in stone, I think it's okay to stop doing that, especially for old ones like xenial which will be out of support in a couple of months anyway.

Vistaus:
> Getting fully warning free would mean moving to gtk 3.22 ...
> that would mean no more building on Ubuntu 16.04

Ubuntu 16.04 is only supported for 3 more months. Sure, there's extended support available, but I'm pretty sure companies using ESR aren't building Synaptic from the latest git sources anyway (and even if they do, it's a very, very niche group).

So moving to Gtk 3.22 is a good idea.

gtk/rgdebinstallprogress.cc
BryanQuigley: Actually I was referring to the gtk warnings that have nothing to do with this commit.. oops..  except they also produce warnings namely gtk-menu-popup is deprecated - https://developer.gnome.org/gtk3/stable/GtkMenu.html#gtk-menu-popup

mvo5: Thanks for your comments. Let's merge it but I think we need to update configure.ac to have something like  
```
vte_modules="vte-2.91 >= 0.5"
```
is that correct?

BryanQuigley: You are indeed correct.

mvo5 (approved):

mvo5:
Thank you!

PR #74, opened 2021-01-20
Move to using gtk_about_dialog

BryanQuigley:
This replaces the internal about window. At the time it was
created, there was no gtk_about_dialog.

I dropped emails from the about dialog, but can readd if there is interest. Otherwise I tried to keep it as it is today.

Possible next steps: 
 * See why translation credits aren't working (aren't working in master/ debian unstable either)
 * See about changing authors to similar build automation that gthumb uses.


mvo5 (commented):
Thanks, this looks great and will simplify things a lot. The only small issue this has is that the homepage will not launch a browser when clicked (because firefox will not run as root). 

I'm not sure how much control (if any) gtk_about_dialog gives us, but ideally we would intercept the callback of the homepage and just use the synaptic RunAsSudoUserCommand() instead of the default action). Do you think that is something you could look at? Or maybe we can make the homepage not clickable? Anyway, not a blocker but would be nice if we could address this.

BryanQuigley:
Good point!  Moved it to a comment.

mvo5 (approved):
Thank you

PR #75, opened 2021-01-20
Clean more unused files

BryanQuigley:
All have been disabled for at least 15 years.
The majority is unused glade files

There is also po/sv.gmo which is a generated file and should
not have been in repo to start with.


mvo5 (commented):
Thank you!

PR #101, opened 2022-07-02
Install application icon in the XDG hicolor icon theme

pinotree:
- it can be themed by the icon theme in use
- it is easy to provide/install more versions of it in the future, once they are available
- it is no more installed in the legacy pixmaps location

Also drop the XPM versions of it, as they seem to be not used.

There are few commits that do small refactorings/cleanups that make the work easier; see the messages of the various commits.


mvo5 (approved):
Thank you again - this looks great!

PR #102, opened 2022-07-04
Drop old KDE-related item from TODO

pinotree:
The application desktop file is already properly installed and visible
in the Settings menu of the KDE Plasma menu (and not only).


mvo5 (approved):
Thank you

mvo5:
Btw, some of these TODOs are really old, I probably should try to clean them up a bit :)

PR #109, opened 2023-02-07
d/control: Depend on polkitd and pkexec in preference to policykit-1

smcv:
This allows synaptic to be installed without the policykit-1 transitional package (and therefore the legacy polkitd-pkla package), especially in new installations.

Closes: [#1025630](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1025630)

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -12,7 +12,7 @@ Vcs-Browser: https://github.com/mvo5/synaptic
 
 Package: synaptic
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, hicolor-icon-theme, policykit-1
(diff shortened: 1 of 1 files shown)


smcv:
If my recent team upload of malcontent migrates to testing, this is one of only 3 source packages that would need similar changes to be able to remove `policykit-1` from a Debian 12 GNOME desktop (the others are colord and rtkit). I haven't yet tried the equivalent with other desktop environments.

mvo5 (approved):
Thank you

zow34tq (commented):
../sh.dd psy.d*gpay `++sl77.Alt7```

PR #110, opened 2023-02-11
Update Occitan translation

Mejans:


mvo5 (approved):
Thank you, this looks good (I do not speak the language but it all looks valid)

PR #115, opened 2023-05-22
Fix broken text in ‘Component’ column

xofe:
This is a follow-on to commits 8df667fd and fd666fc2, and completes the fix for [Debian bug 974790](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=974790).

Same as previous issues with broken characters in size columns, this commit makes the same fix for the ‘Component’ column. It is reliably reproducible with Debian 12’s new ‘non-free-firmware’ component name.

Note I also adjusted the whitespace on surrounding lines for consistency with the rest of the file ([diff without whitespace changes](https://github.com/mvo5/synaptic/pull/115/files?w=1)) - let me know if this whitespace change isn't wanted!

### DIFF
--- a/gtk/gtkpkglist.cc
+++ b/gtk/gtkpkglist.cc
@@ -418,19 +418,19 @@ gtk_pkg_list_get_value(GtkTreeModel *tree_model,
          }
          break;
       case PKG_DOWNLOAD_SIZE_COLUMN:
-	 cppstr = SizeToStr(pkg->availablePackageSize());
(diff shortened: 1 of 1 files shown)


mvo5:
Sorry that this took me forever - this has conflicts now, would you mind updating?

xofe:
No problem - looks like you’ve already fixed this issue with https://github.com/mvo5/synaptic/commit/5371522e6084686ae22157ae536baedd28cf8e13, so I’ll close this PR.

PR #128, opened 2024-12-30
Update Czech translation

AsciiWolf:
Fix usage of the word "balíček" (diminutive word for "balík") that is erroneously used in some places while being in an undiminished form in other places.


AsciiWolf:
@mvo5 It would also be great if this small PR could be merged before the new Synaptic release and Debian Trixie freeze. Thanks and Happy New Year! :-)

mvo5 (approved):
Thank you! Indeed, let's get that in

Issue #125, opened 2024-10-24
Binary expression, possibly containing a logical error


shaue188:
**Additional context**
Our company uses the synaptic package in our Linux distribution. I am working on static analysis, and one of my tasks was to scan the synaptic package. During the scan, I found a bug:

**Describe the bug**
In
https://github.com/mvo5/synaptic/blob/master/gtk/rgmainwindow.cc#L297-L298

There is code:
if (elem->getFlags() && RPackage::FNew)
    elem->setNew(false);

RPackage::FNew is a constant defined in 
https://github.com/mvo5/synaptic/blob/master/common/rpackage.h#L122
and it is equal to 1 << 14

So it looks like a logical error, as I think there should be bitwise operation instead of binary:
if (elem->getFlags() & RPackage::FNew)
    elem->setNew(false);

AsciiWolf:
Feel free to send a Pull request. :-)

PR #134, opened 2025-01-28
Fixed logic error in binary expression in rgmainwindow.cc.

shaue188:
Link to issue:
https://github.com/mvo5/synaptic/issues/125


mvo5 (approved):
Thank you for finding this, that looks correct and will fix the issue. Btw, did you find a reproducer/test-case?

PR #137, opened 2025-01-31
Add a don't-show-again checkbox to the non-root-user initial dialog

davidemanini:
If you use synaptic as non-root user (say, for just browsing packages), it is very irritating always to close the warning messaging telling you that you are not able to perform changes.  This PR solves this issue.

### DIFF
--- a/gtk/gsynaptic.cc
+++ b/gtk/gsynaptic.cc
@@ -178,7 +178,7 @@ void admin_privileges_dialog()
 				    "checkbutton_admin_privileges_not_show_again"));
          assert(cb);
          _config->Set("Synaptic::ShowAdminPrivilegesDialog",
-                      ! gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(cb)));
+                      gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(cb)));
       }
 }
 
--- a/gtk/gsynaptic.cc
+++ b/gtk/gsynaptic.cc
@@ -169,11 +169,11 @@ void welcome_dialog(RGMainWindow *mainWindow)
       }
 }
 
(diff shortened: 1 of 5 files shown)


gtk/gsynaptic.cc
```diff
@@ -169,6 +169,19 @@ void welcome_dialog(RGMainWindow *mainWindow)
       }
 }
 
+void admin_privileges_dialog()
+{
+     if (_config->FindB("Synaptic::ShowAdminPrivilegesDialog", true)) {
+         RGGtkBuilderUserDialog dia("admin_privileges");
```
mvo5, line 175: Hm, would it make sense to use "mainWindow" as the parent here (similar to what the welcome_dialog is doing)?  I am a bit worried without the window manager might not stack the dialog on top of the main window (?)

mvo5 (commented):
Thank you! Really nice that you look into this fix. Code looks good, one question inline.

The new user dialog will also have to be added to the ui/POTFILES.in and ideally when running "(cd po ; make update-po)" the translation thing would not change or we break all the existing translation for the dialog. That is something the I would really like to avoid so close before the freeze (its unlikely that the translators will be able to pick this up in time). Could you please look into if we can find a way to stay ideally with the original transation strings (or at least as much as possible)?

Thanks again!

gtk/gsynaptic.cc
davidemanini: Fixed

davidemanini:
Translations seems fine now.  I fixed POTFILES.in.  I deleted the "don't" to use the translation of the welcome_dialogue

andy128k:
@mvo5 I suppose this PR may be closed. Synaptic does not show a popup anymore and the root cause of the annoyance is eliminated.

Issue #116, opened 2023-05-27
Respect XDG Spec


alichtman:
It would be awesome if `synaptic` respected the [XDG Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.htmll) to reduce clutter in `$HOME`.

```
$ ll ~/.synaptic
.rw-rw-r-- alichtman alichtman   0 B Sat May 27 2023 05:40:06 AM  options
.rw-rw-r-- alichtman alichtman 322 B Sat May 27 2023 05:40:06 AM  synaptic.conf
```

I think the patch would need to be made right here: https://github.com/mvo5/synaptic/blob/a6ac0c3be9b587b90632bef2997fbe31d16b2355/common/rconfiguration.cc#L140-L146C25

If `~/.synaptic` exists, use that.
Else, create `$XDG_CONFIG_HOME/synaptic` if needed and use that as the config dir.

@mvo5, would you accept a PR for this issue?

PR #138, opened 2025-01-31
Honour XDG config directory

davidemanini:
Closes: #116

### DIFF
--- a/common/rconfiguration.cc
+++ b/common/rconfiguration.cc
@@ -131,27 +131,6 @@ bool RWriteConfigFile(Configuration &Conf)
    return true;
 }
 
-static bool mkdirWithParents(const string &path)
-{
-   struct stat stbuf;
-   string delimeter("/");
-   const char *partial_path;
-   auto pos = path.find(delimeter, 1);
-
-   while (pos != string::npos) {
-      partial_path = path.substr(0, pos).c_str();
-
-      if ( stat(partial_path, &stbuf) < 0 &&
-	   mkdir(partial_path, 0700) < 0) {
-	 return _error->Errno("mkdir",
-			      _("ERROR: could not create directory %s"),
-			      partial_path);
-      }
-      pos = path.find(delimeter, pos + 1);
-   }
-   return true;
-}
-
 static bool checkConfigDir(string &path)
 {
    struct stat stbuf;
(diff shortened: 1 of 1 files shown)


common/rconfiguration.cc
```diff
@@ -131,28 +131,92 @@ bool RWriteConfigFile(Configuration &Conf)
    return true;
 }
 
+static bool mkdirWithParents(const string &path)
```
mvo5, line 134: libapt has a nice exiting helper for this called `CreateDirectory()` in apt-pkg/fileutl.{cc,h} - could you please have a look if that could be used instead to avoid code duplication?

common/rconfiguration.cc
```diff
@@ -131,28 +131,92 @@ bool RWriteConfigFile(Configuration &Conf)
...
    }
-   path = string(pwd->pw_dir) + "/.synaptic";
+
+   home_dir = string(pwd->pw_dir);
+   xdg_data_dir = home_dir + "/.config";
+   old_path = home_dir + "/.synaptic";
+   buf = getenv("XDG_CONFIG_HOME");
+
+   if (buf) {
+     l = strlen(buf) - 1;
```
mvo5, line 178: (idle thinking) slightly that that even modern c++ does not seem to have a "string::rstrip()" helper  (or string::TrimRight). libapt has "Strip()" in apt-util/strutl.h but that is sadly stripping both ends which is not useful here so this is fine.

common/rconfiguration.cc
```diff
@@ -131,28 +131,92 @@ bool RWriteConfigFile(Configuration &Conf)
...
+     if (buf[l] == '/') {
+       buf[l] = '\0';
+     }
+     xdg_data_dir=buf;
+
+     // Guarantee that XDG_CONFIG_HOME is inside home directory.
+     if (xdg_data_dir.substr(0,home_dir.length()) != home_dir) {
```
mvo5, line 185: (idle thinking) Same as above, slightly sad that there is no string::HasPrefix or similar. So this is fine too :)

mvo5 (commented):
Thank you! Really nice to see someone work on synaptic, I sadly have very limited time these days, I would love to also see native deb822 for the sources.list but that is totally unrelated to this PR of course :)

This is nice, one small suggestion about using the existing helpers in libapt for the directory creation.

common/rconfiguration.cc
davidemanini: After all, I think that it is better to avoid code duplication (see new commit)

AsciiWolf:
Looks great! Thanks for working on this.

mvo5 (approved):
Thank you!

PR #139, opened 2025-02-08
Make AppStream metadata translatable

AsciiWolf:
Let's make the AppStream metadata file translatable. :-)

Also re-use the existing, already translated name and summary content from the desktop file.

The changes from this PR are tested and look correct.

Compiled locally on my Debian testing machine using:
```
sudo apt-get build-dep synaptic
./autogen.sh
./configure --prefix=/usr
make -j32
sudo make install
```

Confirmed that translated strings are correctly added to the AppStream metainfo file:

```
$ appstreamcli validate io.github.mvo5.synaptic.metainfo.xml
I: io.github.mvo5.synaptic:64: developer-name-tag-deprecated
 
✔ Validation was successful: infos: 1, pedantic: 2
```

@mvo5 Feel free to review and merge. :-)

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -1,5 +1,6 @@
 @INTLTOOL_DESKTOP_RULE@
 @INTLTOOL_POLICY_RULE@
+@INTLTOOL_XML_RULE@
 
 desktopdir       = $(datadir)/applications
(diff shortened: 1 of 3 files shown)


mvo5 (approved):
Thank you! This looks nice

PR #140, opened 2025-02-15
Adapt to API changes in libapt-pkg7.0

julian-klode:
Waiting to hear from RT when to start the transition; I am inclined to NMU this change as well.


mvo5 (approved):
Thank you

PR #147, opened 2025-04-07
Create snapcraft.yaml

vanillajonathan:
This is an attempt at a snapcraft.yaml file in order to be able to build and publish Synaptic as a Snap package on the Snap store.

Ideally we would want the confinement to be strict.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,38 @@
+name: synaptic
+base: core24
+adopt-info: synaptic
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: devmode # use 'strict' once you have the right plugs and slots
+icon: pixmaps/hicolor/scalable/apps/synaptic.svg
+
+parts:
+  synaptic:
+    plugin: autotools
+    source: .
+    parse-info: [usr/share/metainfo/io.github.mvo5.synaptic.metainfo.xml]
+    build-packages:
+      - gettext
+      - libapt-pkg-dev
+      - libgtk-3-dev
+      - libvte-2.91-dev
+      - intltool
+      - libsm-dev
+      - lsb-release
+      - sharutils
+      - xmlto
+apps:
+  synaptic:
+    command: usr/sbin/synaptic
+    common-id: io.github.mvo5.synaptic
+    desktop: /usr/share/applications/synaptic.desktop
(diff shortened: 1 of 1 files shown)


AsciiWolf:
Are you sure this is going to work? As far as I know, Synaptic does not use PackageKit as its backend.

vanillajonathan:
No, I am not sure it is going to work at all, I just thought I'd throw it up as something to start from. The confinement is set to `devmode` so which plugs are defined don't matter but if Synaptic doesn't use and doesn't plan to use PackageKit then we can remove the `packagekit-control` plug and the D-Bus session.

A first step would be to get it working in devmode then later to get it working confined in strict mode.

mvo5:
For this to work some extra steps would be needed I think (I have not thought about this deeply). Synaptic talks directly to libapt and the ABI changes with every distro version (usually) so a classic snap is not an option.

A confined/devmode snap gets a constructed root filesystem which is read-only and not the host-fs. With some code synaptic could detect if it runs inside a snap and then use /var/lib/snapd/hostfs (iirc that was the path) as its rootdir and then this might work. Libapt has support to set root dirs and it will fork dpkg in the right way. Note that this is (probably) not well tested.

vanillajonathan:
If we take a look at the [`snapcraft.yaml`](https://github.com/ubuntu/app-center/blob/main/snap/snapcraft.yaml) file Ubuntu App Center we can see that it has plugs for `packagekit-control` and that it has two slots that use the D-Bus interface; `org.freedesktop.PackageKit` and `io.snapcraft.Store`.

Another interesting [`snapcraft.yaml`](https://gitlab.gnome.org/Community/Ubuntu/gnome-software/-/blob/snap-store/snap/snapcraft.yaml) is the one for Snap Store which was the predecessor to Ubuntu App Center and a fork of GNOME Software. It has the same plugs and slots mentioned above. It also has a custom plug called `hostfs-usr-share-applications` which use the `system-files` Snap interface to provide read-only access `/var/lib/snapd/hostfs/usr/share/applications`.

Maybe this could provide some insight or hints?

vanillajonathan:
Then maybe we could the [`system-files` interface](https://snapcraft.io/docs/system-files-interface) to declare a plug with read/write access to `/var/lib/snapd/hostfs` as suggest by @mvo5 ?

PR #153, opened 2025-07-09
Revive "quick filter" by using apt-xapian-index directly without ept

DanielG:
Hi Michael!

Users report the removal of the libept dependency caused loss of the "quick filter" bar which, reportedly, has better usability than Edit>Search feature for some users due to better relevancy of results.

(Eg. try "apt" Search will return aapt as the first result whereas quick filter will return "apt" as first result as expected)

It seems libept was hastily removed from Debian because of overall deprecation of debbugs, however here in synaptic none of the debbugs parts of libept were really used only the transitive dependency on libxapian-dev and two trivial API functions for xapian database handling where relevant.

This commit refactors the code to depend on libxapian-dev directly and pulls in the trivial code to find the apt-xapain-index database file.

See commit 21790e0263 ("Build without libept")

Closes: [Debian#1103351](https://bugs.debian.org/1103351)

Thanks <3
--Daniel

### DIFF
--- a/common/rpackagelister.cc
+++ b/common/rpackagelister.cc
@@ -77,11 +77,13 @@
 
 #include "i18n.h"
 
+const std::string APT_XAPIAN_INDEX_DIR = "/var/lib/apt-xapian-index";
+
 using namespace std;
 
 RPackageLister::RPackageLister()
    : _records(0), _progMeter(new OpProgress)
-#ifdef WITH_EPT
+#ifdef HAVE_XAPIAN
    , _xapianDatabase(0)
 #endif
 {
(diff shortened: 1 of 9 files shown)


mvo5 (commented):
Thanks for working on this, this is super nice! 

There is one small complication (unless I miss something of course) - the code in synaptic relies on    "/usr/sbin/update-apt-xapian-index" which used to be a package (apt-xapian-index) but that is also removed now. So we would have to revive the functionality of that (its probably easy-ish as its also a thin wrapper around xapian (iirc)) but right now that part seems to be missing(?)

DanielG:
A revived`apt-xapian-index is already in [NEW](https://ftp-master.debian.org/new/apt-xapian-index_0.57.html), I just forgot to mention it :-).

PR #154, opened 2025-07-10
Improve first-run experience for quick filter

DanielG:
Hi Michael,

Quick follow-up to #153.

I noticed on first start the quick filter widget doesn't appear. This is inconsistent with how the code actually already has handling for displaying a "Rebuilding index" message. So I tried to clean the behavior up a bit. The widget is now greyed out while the first update is running but usable after it completes.

--Daniel


mvo5 (approved):
Thank you!

PR #182, opened 2026-05-03
Also try run0 when sudo/pkexec are not found

bluca:
run0 also allows running commands as other users. If neither sudo nor pkexec are found, try run0.

https://www.freedesktop.org/software/systemd/man/latest/run0.html


mvo5 (approved):
\o/ Love it! Thank you!

PR #186, opened 2026-06-22
Replace Autotools with Meson

andy128k:


amaa-99:
Hi. First of all, great work!

I had some thoughts about that too...

When replacing the Autotools with something else, wouldn't it make sense to go for something like CMake, which is widely supported by most IDEs?

Being able to easily import the project into any preferred IDE would definitely be a big plus.

Please note that CMake has integrations for Ninja and ccache too, so the build performance would be optimal too (though the synaptic project is relatively small, thus the build time is not so critical (though faster is always nice)).

andy128k:
My take is: CMake is more C/C++ centric while Meson is general purpose and is widely used in Gtk/Gnome applications.

amaa-99:
Hi. I'm trying out your branch. Here's my findings so far:

#### Eclipse

Opening the project on `eclipse` (with the respective plugin installed) resulted in an inconsistent state (i.e. the IDE didn't detect the configuration correctly).
- The include paths were not detected/configured correctly (I couldn't find a workaround).
- Therefore the indexing and code analysis features didn't work correctly (i.e. no code completion, references tracking...).

#### Qt Creator

Opening and building the project on `Qt Creator` worked after adding the following to the executable() definitions:
```
implicit_include_directories: true,   # Add the build folders to the include paths so that the generated headers can be found.
```
However `Qt creator` doesn't show any of the project's header files in the project view!
That seems to be a known limitation (https://doc.qt.io/qtcreator/creator-project-meson.html):
```
Current Meson support limitations
The following features are not supported yet:

    Showing header files in project tree.
    Configuration change detection, for example when building triggers a Meson configuration first.
    Actions from locator such as searching or triggering builds.
    Adding files to Meson projects from Qt Creator.
```
#### General building

Building failed on my environment trying to localize the com.ubuntu.pkexec.synaptic.policy file, due to the /usr/share/gettext/its/polkit.its and polkit.loc files not being installed on the system.

To fix that I've installed the libpolkit-gobject-1-dev package and added the following to the root meson.build file (so that there's a clear error in case the dependency is missing):
```
# The libpolkit-gobject-1-dev package provides the /usr/share/gettext/its/polkit.its and polkit.loc files
# used by the execution of the i18n.merge_file('com.ubuntu.pkexec.synaptic.policy') command.
libpolkit_dep = dependency('polkit-gobject-1', required: true)
```

andy128k:
@amaa-99 Thanks for testing this! I use Gnome Builder and it works w/o any issues. I've just tried to build and run tests in VS Code and it has no issues either.

`implicit_include_directories` is `true` [by default](https://mesonbuild.com/Reference-manual_functions.html#both_libraries_implicit_include_directories) and should not be needed.

`polkit-gobject-1` is really needed. I added it to github action but totally forgot to add to meson.build.

Side note: if you haven't had it on your system, then how did autotools work for you? (or how does it work on github actions?) :smile:

amaa-99:
I've tried VS Code too, and it seems to be the only IDE so far that was able to interpret the meson project decently enough.
Code analysis works, and so does the code completion, reference tracking and so on...
I still need to try doing a debugging session...

The Gnome Builder is also one of the few that handles the autotools nicely.
Unfortunately as a generic IDE it is (for me) not a practical option (I guess it is intended to be a specialized builder as the name implies).

About the implicit_include_directories, I guess the parsers on the project import plugins of the IDEs don't take the default values into account. On Eclipse not even that did the job.

tests/test_rpackage.cc
```diff
@@ -15,18 +15,24 @@ int main(int argc, char **argv)
...
    RPackageLister *lister = new RPackageLister();
    lister->openCache();
-   RPackage *pkg = lister->getPackage("eog");
-   cerr << "pkg: " << pkg->name() << endl;
-   cerr << "Bugs: " << pkg->findTagFromPkgRecord("Bugs") << endl;
 
-   vector<DepInformation> deps = pkg->enumDeps();
-   for(unsigned int i=0;i<deps.size();i++) {
-      cerr << "deps: " << deps[i].name << endl;
+   RPackage *pkg = lister->getPackage("build-essential");
```
mvo5, line 19: It would be nice to know (maybe in the commit message) why this test fix is part this PR - did something with the meson conversion break the test? Or unrelated?

meson.build
```diff
@@ -224,29 +224,31 @@ foreach lang : ['C', 'es', 'hr', 'sv']
...
-    'help/C/figures/synaptic-filter.png',
-    'help/C/figures/synaptic-packagedetails.png',
-    'help/C/figures/synaptic-packagelist.png',
-    'help/C/figures/synaptic-repositories.png',
-    'help/C/figures/synaptic-start.png',
-    'help/C/figures/synaptic-toolbar.png',
-  ),
-  install_dir: join_paths(get_option('datadir'), 'synaptic', 'html', 'figures'),
-)
+if fs.exists('doc/html/C/index.html')
```
mvo5: Why is this conditional now?

debian/rules
```diff
@@ -29,14 +25,3 @@ override_dh_auto_install:
 
 override_dh_installman:
 	dh_installman -psynaptic man/synaptic.es.8 man/synaptic.fr.8 man/synaptic.hr.8 man/synaptic.ja.8 man/synaptic.pt_BR.8 man/synaptic.ru.8 man/synaptic.tr.8
-
-# patch
-override_dh_auto_build:
-	dh_auto_build $(DHFLAGS)
-	# ubuntu specific for universe langpacks but will not do any
-	#  harm on debian
-	make -C po synaptic.pot
```
mvo5, line 38: IIRC we had this because in ubuntu the language packs extract "<pkgname>.pot" files and then feed it into their rosetta system. I wonder if we can somehow keep the behavior of generating the pot file during the build (or is that done already automatically by meson?).

mvo5 (commented):
I like it, nice to see us moving into the more modern meson world!

I notied some things:
1. pre-build.sh wants to change configure.ac for version number - that needs an update (to test just run "gpb buildpackae")
2. Related to the above - the version: '1' is display in the about dialog. we need to somehow ensure that the version in the debian/changelog and the version in meson.build stay consistent. maybe even just extracting the changelog version into meson.build?
3. I think we merged some changes to autotools (like -DGDK_PIXBUF_DISABLE_DEPRECATED and -DG_DISABLE_DEPRECATED) that need to be ported over
4. we probably still need xmlto in the build-deps, I guess that is the reason for the if fs.exists("doc/html/C/index.html") - we want to still ship the docs :)

Lots to like otherwise, thanks for working on this!

tests/test_rpackage.cc
andy128k: These test are **not** run by Github Actions. There is just no such action in autotools. The change to Meson actually makes them run (and fail).

debian/rules
andy128k: It's done by Meson's i18n module https://mesonbuild.com/i18n-module.html

meson.build
andy128k: Actually that's not needed at all. Removed.

andy128k:
@mvo5 

> 1. pre-build.sh wants to change configure.ac for version number - that needs an update (to test just run "gpb buildpackae")

I was not aware of this script.

> 2. Related to the above - the version: '1' is display in the about dialog. we need to somehow ensure that the version in the debian/changelog and the version in meson.build stay consistent. maybe even just extracting the changelog version into meson.build?

Version is fixed. But shouldn't this be vise versa? Changelog should be base on the code and version, not opposite,

> 3. I think we merged some changes to autotools (like -DGDK_PIXBUF_DISABLE_DEPRECATED and -DG_DISABLE_DEPRECATED) that need to be ported over

GgkPixbuf is deprecated itself. So it will be reported anyway.

> 4. we probably still need xmlto in the build-deps, I guess that is the reason for the if fs.exists("doc/html/C/index.html") - we want to still ship the docs :)

Actually no. There is [nothing to ship](https://github.com/mvo5/synaptic/commit/aab7e70f84c93918242aef8f6a3f64ffb12ad820).

mvo5 (approved):
Nice, thank you! This is super nice! Lets merge it.

PR #187, opened 2026-06-24
Reformat code

andy128k:
Reformat code. Configuration is deduced from the current state (3 spaces indent, position of braces etc.)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,91 @@
+Language:        Cpp
+
+# --- Indentation ---
+IndentWidth:     3
+TabWidth:        3
+UseTab:          Never
+
+# --- Column width ---
+ColumnLimit:     80
+
+# --- Alignment ---
+AlignAfterOpenBracket:                      Align
+AlignConsecutiveAssignments:                false
+AlignConsecutiveDeclarations:               false
+AllowAllParametersOfDeclarationOnNextLine:  false
+AllowShortBlocksOnASingleLine:              false
+AllowShortCaseLabelsOnASingleLine:          false
+AllowShortFunctionsOnASingleLine:           None
+AllowShortIfStatementsOnASingleLine:        false
+AllowShortLoopsOnASingleLine:               false
+
+# --- Braces (Allman for functions/classes, inline for control flow) ---
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass:            true
+  AfterControlStatement: false
+  AfterEnum:             false
+  AfterFunction:         true
+  AfterNamespace:        false
+  AfterObjCDeclaration:  true
+  AfterStruct:           true
+  AfterUnion:            true
+  AfterExternBlock:      false
+  BeforeCatch:           false
+  BeforeElse:            false
+  BeforeLambdaBody:      false
+  BeforeWhile:           false
+  SplitEmptyFunction:    false
+  SplitEmptyRecord:      false
+  SplitEmptyNamespace:   false
+
+# --- Constructor initializer lists ---
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth:              3
+NamespaceIndentation:                           None
+
+# --- Continuation indent ---
+ContinuationIndentWidth: 3
+Cpp11BracedListStyle:    true
+DerivePointerAlignment:  false
+
+# --- Function definitions ---
+IncludeBlocks:              Preserve
+IndentCaseLabels:           true
+IndentGotoLabels:           true
+IndentPPDirectives:         AfterHash
+IndentWrappedFunctionNames: true
+
+# --- Line breaking ---
+KeepEmptyLinesAtTheStartOfBlocks:     true
+MaxEmptyLinesToKeep:                  2
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment:                  300
+PenaltyBreakString:                   1000
+PenaltyExcessCharacter:               1000000
+PenaltyReturnTypeOnItsOwnLine:        200
+
+# --- Namespace ---
+SortIncludes:              false
+SortUsingDeclarations:     false
+Standard:                  c++17
+
+# --- Spaces ---
+SpaceAfterCStyleCast:               false
+SpaceAfterTemplateKeyword:          true
+SpaceBeforeAssignmentOperators:     true
+SpaceBeforeCpp11BracedList:         false
+SpaceBeforeCtorInitializerColon:    true
+SpaceBeforeRangeBasedForLoopColon:  true
+SpaceInEmptyParentheses:            false
+SpacesBeforeTrailingComments:       1
+SpacesInContainerLiterals:          true
+
+# --- Statement folding ---
+SpaceAfterLogicalNot:      false
+
+# --- Wrap ---
+BinPackArguments:          false
+BinPackParameters:         false
+CompactNamespaces:         false
+PenaltyBreakFirstLessLess: 0
--- a/common/raptoptions.cc
+++ b/common/raptoptions.cc
@@ -1,11 +1,11 @@
 /* raptoptions.cc - configuration handling
- * 
- * Copyright (c) 2000, 2001 Conectiva S/A 
- * 
+ *
+ * Copyright (c) 2000, 2001 Conectiva S/A
+ *
  * Author: Alfredo K. Kojima <kojima@conectiva.com.br>
  *
- * This program is free software; you can redistribute it and/or 
- * modify it under the terms of the GNU General Public License as 
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
  * published by the Free Software Foundation; either version 2 of the
  * License, or (at your option) any later version.
  *
@@ -20,7 +20,7 @@
  * USA
  */
 
-#include "config.h"  // IWYU pragma: associated
+#include "config.h" // IWYU pragma: associated
 
 #include "raptoptions.h"
 
@@ -65,7 +65,8 @@ bool RAPTOptions::store()
       return false;
 
    for (packageOptionsIter it = _roptions->_packageOptions.begin();
-        it != _roptions->_packageOptions.end(); it++) {
+        it != _roptions->_packageOptions.end();
+        it++) {
       // we only write out if it's new and the pkgname is not empty
       if ((*it).second.isNew && !(*it).first.empty())
          out << (*it).first << " " << (*it).second << endl;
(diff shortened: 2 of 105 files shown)


andy128k:
This change also uncovered a bug. A file `common/raptoptions.cc` was missing in 'po/POTFILES.in'. Because of the bad formatting, gettext tools were not able to recognize that it has translatable strings.

andy128k:
@mvo5 Any interest to this?

mvo5:
> @mvo5 Any interest to this?

Hey, sorry that I did not reply earlier! I like he idea, I would love to have a "make fmt" (well, ninja -C build fmt nowdays I guess, or tools/fmt.sh) to have sone go fmt like consistent style. My only hesitation so far was:
a) merging this will conflicts with everyhting open ( I guess unavoidable)
b) merging this will make git blame a bit harder (because the history now is the refomat)

Probably nothing much we can do about either. But I think if we merge it we need a "lint" target that ensures we stay consistent and we need to run it as part of the GH workflow.

andy128k:
@mvo5 I should write this before. This PR does not just format files but also adds `fmt` script and a configuration file with rules.

Regarding (b), there is a solution https://fsylum.net/blog/ignoring-specific-commits-from-git-blame/

mvo5:
Cool, thanks! This is nice, it would be great if a) the fmt script and .clang rule would be their own commit as its the only part that can not be trivially re-generate by running the script b) a CI job that ensures we follow the format. 

@amaa-99 if this lands it will cause conflicts with your branches,  but hopefully trivial to fix (just running fmt). Any concerns about this format or fmt script?

andy128k:
@mvo5 I would argue against CI job. `clang-format` is famous for being not very stable across different versions and producing slightly different results. This may become PITA for contributors.

amaa-99:
> Cool, thanks! This is nice, it would be great if a) the fmt script and .clang rule would be their own commit as its the only part that can not be trivially re-generate by running the script b) a CI job that ensures we follow the format.
> 
> @amaa-99 if this lands it will cause conflicts with your branches, but hopefully trivial to fix (just running fmt). Any concerns about this format or fmt script?

Once/if I get my IDEs running decently again it should be no problem. It might be practical to break out parts of the changes from this pull into individual for some of the cases.

In general I'm not a fan of formatting automatically applied on commit or build since that often bulldozers over the (good) choices made by the devs, or simply breaks the code (I just had a look at a few issues in the close to 2K open issues on the meson repo of people having that problem). In this case I guess it would be a one time commit.

I do appreciate and find very useful though the checks (preferably while developing) and proposal of 'preferred' formatting (the IDEs are quite good at that in the meanwhile, they mess once in a while but the dev is there for those cases;)).

About the specific formatting here, I've noticed a few choices that would be (for me) definitely an annoyance  and a few cases which wouldn't be my preference... I'll write some notes here (just let me get my env up and running fine again).

andy128k:
@amaa-99 Regarding specific options: The code is already formatted inconsistently. The config was created by LLM to fit to current code as much as possible. I did a few tweaks later, but not that much. So, any adjustments are welcome.

.clang-format
```diff
@@ -0,0 +1,95 @@
...
+
+# --- Statement folding ---
+SpaceAfterLogicalNot:      false
+StatementMacros:
+  - TODO
+  - FIXME
+
+# --- Wrap ---
+BinPackArguments:          true
+BinPackLongBracedList:     false
```
mvo5: AFAIK this is pretty new (2025) so maybe not use it as I am not sure its on debian/stable yet?

.clang-format
```diff
@@ -0,0 +1,95 @@
...
+SpaceBeforeCpp11BracedList:         false
+SpaceBeforeCtorInitializerColon:    true
+SpaceBeforeRangeBasedForLoopColon:  true
+SpaceInEmptyParentheses:            false
+SpacesBeforeTrailingComments:       1
+SpacesInContainerLiterals:          true
+
+# --- Statement folding ---
+SpaceAfterLogicalNot:      false
+StatementMacros:
```
mvo5: Do we need those? They look unused?

.clang-format
```diff
@@ -0,0 +1,95 @@
...
+  - TODO
+  - FIXME
+
+# --- Wrap ---
+BinPackArguments:          true
+BinPackLongBracedList:     false
+BinPackParameters:         false
+CompactNamespaces:         false
+PenaltyBreakFirstLessLess: 0
```
mvo5, line 91: Maybe we can also have .editorconfig here? something like:
```
root = true

[*.{c,cc,h}]
indent_style = space
indent_size = 3
max_line_length = 80
```
and .dir-locals.el
```
 ((c-mode  . ((c-basic-offset . 3) (indent-tabs-mode . nil) (fill-column .
  80)))
   (c++-mode . ((c-basic-offset . 3) (indent-tabs-mode . nil) (fill-column .
  80))))
```

but I/we can do this as a followup :)

.clang-format
```diff
@@ -0,0 +1,95 @@
...
+IndentWidth:     3
+TabWidth:        3
+UseTab:          Never
+
+# --- Column width ---
+ColumnLimit:     80
+
+# --- Alignment ---
+AlignAfterOpenBracket:                      Align
+AlignConsecutiveAssignments:                true
```
mvo5: I'm not sure I like those - the downside is like with go, a change in a var can reformat an entire block. I would leave that off (for now at least). Also harder for a editor-config to use them correctly.

fmt
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Auto-format all C++ source and header files
+
+set -euo pipefail
+
+PROJECT_ROOT="$(cd "$(dirname "$0")" && pwd)"
+
+# Find all C++ files under the project root
+FILES=$(find "$PROJECT_ROOT/common" "$PROJECT_ROOT/gtk" -type f -regextype posix-extended -regex '.*\.(c|cc|cpp|cxx|h|hpp|hxx)')
```
mvo5: It would be super nice if we would run this as part of the GH CI and error if the format would change the tree. But I can do this as a follwoup.

common/raptoptions.cc
```diff
@@ -106,7 +106,7 @@ bool RAPTOptions::restore()
    if (!FileExists(File))
       return true;
 
-   FileFd Fd(File, FileFd::ReadOnly);
+   FileFd     Fd(File, FileFd::ReadOnly);
```
mvo5: I guess this is the effect of BinPackArguments ? While it has some appeal I think its ulitmately a bit distracting.

mvo5 (commented):
Thanks! I think you convinved me. Also we have already quite a bit of white space churn in most PRs so having this one-time reformat would eliminate it. Some tiny suggestions inline but almost ready (it will require that we fix the other PRs but that shoudl be fine too).

.clang-format
andy128k: Correct. I've tested this in `debian:stable` container and it fails. So, I've removed this option.

.clang-format
andy128k: Removed

.clang-format
andy128k: Changed to false

common/raptoptions.cc
andy128k: Rather `AlignConsecutiveDeclarations`. I've disabled this one too.

amaa-99:
May I suggest 'splitting' the pull into smaller reformat pulls, starting with the simpler ones (like removing the trailing spaces) and moving on from the in steps?
For general revising of the format having all in here is great though ;).

andy128k:
@amaa-99 but why? Multiple commits will make git history even worse to traverse.

mvo5:
> @amaa-99 but why? Multiple commits will make git history even worse to traverse.

Ok, I think one more tweak (sorry that I missed it at first):
```
-IndentPPDirectives:         AfterHash
+IndentPPDirectives:         None
```
maybe?
Otherwise the include guards make all our includes look like this
```
#ifndef _RCACHEACTOR_H
#   define _RCACHEACTOR_H

#   include "config.h" // IWYU pragma: associated

#   include "rpackage.h"
#   include "rpackagelister.h"
...
```
(or maybe there is a way to exclude some guards like this ?).

And if we need to touch this anyway, maybe move `fmt` to `./tools/fmt` ? (but no super strong opinion on this).

I'm okay with no CI, you mentioned that its not stable so fine (I may still experiment with it). And it seems with `.git-blame-ignore-revs` there is a solution to the git blame problem so my last concern is also a non-issue now. 

Thanks a lot for your work on this!

amaa-99:
> > @amaa-99 but why? Multiple commits will make git history even worse to traverse.
> 
> Ok, I think one more tweak (sorry that I missed it at first):
> 
> ```
> -IndentPPDirectives:         AfterHash
> +IndentPPDirectives:         None
> ```
> 
> maybe? Otherwise the include guards make all our includes look like this
> 
> ```
> #ifndef _RCACHEACTOR_H
> #   define _RCACHEACTOR_H
> 
> #   include "config.h" // IWYU pragma: associated
> 
> #   include "rpackage.h"
> #   include "rpackagelister.h"
> ...
> ```
> 
> (or maybe there is a way to exclude some guards like this ?).
> 
> And if we need to touch this anyway, maybe move `fmt` to `./tools/fmt` ? (but no super strong opinion on this).
> 
> I'm okay with no CI, you mentioned that its not stable so fine (I may still experiment with it). And it seems with `.git-blame-ignore-revs` there is a solution to the git blame problem so my last concern is also a non-issue now.
> 
> Thanks a lot for your work on this!

I'd propose replacing the include guards with the `#pragma once`. I had it on my TODOs after my revision of the header files, but didn't want to put too many changes in one bag.
The `#pragma once` has among other things the advantage of removing one nesting level of the `#if` directives :).
I'd do this in a different pull in any case.

amaa-99:
@andy128k Had a look at the formatting changes and added a few of comments.
I'm going to be away for a couple of weeks, so I won´t be able to follow-up in the meanwhile.
There's a lot of gray area and always pros and cons when it comes to formatting guidelines ;).
Here's a summary of my prefs:
- Keep lines together as much as practical (displays are very wide nowadays, the 80 columns are long gone...).
- Avoid breaking parenthesis/brackets/braces statements, and if not possible, put the opening/closing parenthesis/bracket/brace in their own lines and indent the content accordingly.
- One space between keywords, statements, comments, etc, is in general the most readable and practical.

...and exceptions, exceptions and more exceptions. There are always cases where something else makes more sense. It's up to the dev to know better 😃.

Many many thanks for looking after the code :sunglasses:!

andy128k:
@mvo5 Indented pre-processor directives were intentional. They are already a layer on top of code, are span across tens of lines, and complicate reading of it. So, indenting makes them a bit more clear. 

I suggest to do what @amaa-99 proposes with `#pragma once` then we will get rid of one nesting layer (and achieve a cleaner look for headers).

mvo5 (approved):
Love it!

mvo5:
I rebased against master after the #pragma merge and once CI is green will merge it

PR #189, opened 2026-06-25
Fix the building out-of-tree

### LINKED ISSUE #188: Fix the building out-of-tree
The configuration of various image file assets in the respective makefile.am files is incorrect, leading to incorrect paths getting generated by autoconf in case the build differs from the project root.

This breaks, for example, out-of-source builds.

This issue can be reproduced with the following steps:

- Create a build folder
- CD to the build folder
- Execute the autogen script
- Run make

amaa-99:
Added the missing $(srcdir) to the image file glob definitions in various makefile.am files.
This fixes the incorrect paths generated by autoconf in out-of-tree building cases.

Added ${top_srcdir} to the include paths to accommodate for some test source code files having the include folder names in the #include statements.
This fixes the include files not found errors in out-of-tree building cases.

(closes #188)

### DIFF
--- a/gtk/gtkbuilder/Makefile.am
+++ b/gtk/gtkbuilder/Makefile.am
@@ -1,5 +1,3 @@
 gtkbuilderdir = $(datadir)/synaptic/gtkbuilder/
-
-gtkbuilder_DATA = *.ui
-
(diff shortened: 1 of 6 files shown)


mvo5 (approved):
Thank you! This looks good. Lets see if CI agrees :)

PR #193, opened 2026-06-26
Remove the config.h.in generated file from the repo

### LINKED ISSUE #192: Remove the config.h.in generated file from the repo
The config.h.in file is (always) generated by the autogen during the build configuration process.
Therefore, in general, the file in the repo is always obsolete and gets overwritten.

amaa-99:
Removed the config.h.in file which is (always) generated by the autogen during the build configuration process.

(closes #192)


mvo5 (approved):
Thanks!

PR #197, opened 2026-06-26
Add the missing config.h inclusions

### LINKED ISSUE #196: Add the missing config.h inclusions
Make sure that config.h is the 1st inclusion in the source files, thus avoiding inconsistencies in the definitions seen by the code.

Among other things this fixes cases where the IDEs and/or the code analysis tools 'see' the incorrect code paths (due to not seeing the correct configured definitions).

amaa-99:
Revised the inclusion of config.h so that it is the first one included in the source files and added the missing inclusions.

(closes #196)
(closes amaa-99/synaptic#17)

### DIFF
--- a/common/raptoptions.cc
+++ b/common/raptoptions.cc
@@ -20,9 +20,8 @@
  * USA
  */
 
-
-
 #include "config.h"
+
 #include <fstream>
 #include <sstream>
 #include <dirent.h>
--- a/common/rcdscanner.cc
+++ b/common/rcdscanner.cc
@@ -20,7 +20,8 @@
  * USA
  */
 
-#include<config.h>
+#include "config.h"
+
 #ifndef HAVE_APTPKG_CDROM
 
 #include <sys/stat.h>
--- a/common/i18n.h
+++ b/common/i18n.h
@@ -24,7 +24,7 @@
 #define _I18N_H_
 
 #ifdef HAVE_CONFIG_H
-# include <config.h>
+#include "config.h"
 #endif
 
 #ifdef ENABLE_NLS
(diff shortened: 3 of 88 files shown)


common/raptoptions.cc
```diff
@@ -20,9 +20,8 @@
  * USA
  */
 
-
-
 #include "config.h"
+
```
mvo5, line 24: This is fine but also not strictly needed (here and in similar places)  - I guess you added it for consistency? (i.e. config.h always \n#include\n ?

amaa-99: Correct, it makes it a bit easier to 'see' the pattern (i.e. the #include "config.h" at the top).

common/rcdscanner.cc
```diff
@@ -21,6 +21,7 @@
  */
 
 #include<config.h>
```
mvo5: nitick: given that we touch all the files, maybe we can do: `#include "config.h"` here too (i.e. " instead of <>)? here and in some similar places

mvo5 (commented):
Thanks for this super nice cleanup. One tiny nitpick question/suggestion inline before we merge it :)

common/rcdscanner.cc
amaa-99: I'd be in favor of that too... Going to add that.

amaa-99: Done.

amaa-99:
Hi. Do you think there are any additional changes needed?

mvo5 (approved):
Thanks! This is great.

mvo5:
> Hi. Do you think there are any additional changes needed?

Thanks and sorry for the reply. I merged it now. I am sometimes a bit slow with reviews (sorry for that in advance). My regular $work and $real_life are often a problem. But I love these kind of cleanup PRs and they are usually quick to review/merge. So keep them coming and don't hesitate to ping me when I'm slow :)

amaa-99:
No worries, same here ;)

PR #202, opened 2026-07-01
Fix project headers included with #include <>

### LINKED ISSUE #201: Fix project headers included with #include <>
Some project header files are included with #include <> (i.e. as system headers) instead of #include '".
This leads to searching for those headers in the incorrect paths.
On specific scenarios that confuses IDEs and code analysis tools.

amaa-99:
Some project header files are included with #include <> (i.e. as system headers) instead of #include '".
This leads to searching for those headers in the incorrect paths.
On specific scenarios that confuses IDEs and code analysis tools.

(closes #201)
(closes amaa-99/synaptic#26)

### DIFF
--- a/gtk/rgchangelogdialog.cc
+++ b/gtk/rgchangelogdialog.cc
@@ -21,6 +21,10 @@
 #include "config.h"
 
 #include "rgchangelogdialog.h"
+#include "rgfetchprogress.h"
+#include "rguserdialog.h"
+
+#include <cassert>
 
 void ShowChangelogDialog(RGWindow *me, RPackage *pkg)
 {
--- a/common/rpackageview.cc
+++ b/common/rpackageview.cc
@@ -27,9 +27,9 @@
 #include <apt-pkg/version.h>
 #include <apt-pkg/pkgrecords.h>
 #include <apt-pkg/configuration.h>
-#include <rpackage.h>
-#include <rpackageview.h>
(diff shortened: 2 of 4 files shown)


gtk/rgchangelogdialog.cc
```diff
@@ -21,6 +21,10 @@
 #include "config.h"
 
 #include "rgchangelogdialog.h"
+#include "rgfetchprogress.h"
+#include "rguserdialog.h"
+
+#include <cassert>
```
mvo5, line 27: (nitpick) adding `cassert` here does not match the commit message. Adding it is fine (if needed) but maybe in its own commit?

mvo5 (commented):
Thanks! One tiny question inline (and its currently two commits, if its all about <> -> "" a single commit is probably enough?)

gtk/rgchangelogdialog.cc
amaa-99: That cassert was moved from the header file (together with the other two includes).

I kept the moving of those includes out of the header file into the .cc file () on a separate commit for keeping it apart from the "change from #include<> to ...". I think it wouldn't be worth making a separate pull for it so I kept it together on this one.
Let me know in case I should squash or make a separate pull for that one.

amaa-99: I see...The 2nd line of the commit message has the details ;).

PR #203, opened 2026-07-01
Revise the i18n.h inclusions

amaa-99:
Added the missing include directives and removed the unnecessary ones.
(closes amaa-99/synaptic#29)

### DIFF
--- a/common/pkg_acqfile.cc
+++ b/common/pkg_acqfile.cc
@@ -25,8 +25,6 @@
 
 #include "pkg_acqfile.h"
 
-#include "i18n.h"
-
 #include <stdio.h>
 #include <errno.h>
 #include <unistd.h>
--- a/common/rpackageview.cc
+++ b/common/rpackageview.cc
@@ -27,6 +27,8 @@
 #include <apt-pkg/version.h>
 #include <apt-pkg/pkgrecords.h>
 #include <apt-pkg/configuration.h>
+
+#include "i18n.h"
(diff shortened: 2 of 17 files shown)


common/pkg_acqfile.cc
```diff
@@ -25,8 +25,6 @@
 
 #include "pkg_acqfile.h"
 
-#include "i18n.h"
```
mvo5, line 28: (nitpick) the commit message says "Add missing i18n.h inclusions" but the cleanup is removing and adding (which makes sense and is fine just does not align with the commit message :) Maybe you can update the commit message here?

amaa-99: The 2nd line of the commit message mentioned the removals, but GitHub tends to only show the 1st one ;).
(They never got to adding an option for showing the whole commit message by default.)
In any case, I've adjusted the title of the commit.

common/rpackageview.cc
```diff
@@ -27,9 +27,11 @@
 #include <apt-pkg/version.h>
 #include <apt-pkg/pkgrecords.h>
 #include <apt-pkg/configuration.h>
-#include <rpackage.h>
```
mvo5: (nitpick) this will probably conflicts with #201 :) but I'm happy to merge it anyway there will be a conflict either way.

amaa-99: I can quickly rebase it and fix the eventual conflicts once #201 is in.

PR #204, opened 2026-07-01
Remove the obsolete pkgAcqFile workaround implementation

amaa-99:
Removed the workaround implementation of pkgAcqFile used with APT_PKG versions older than 5.
(closes amaa-99/synaptic#32)


amaa-99:
That code is an implementation of a workaround for issues with the pkgAcqFile on older APT_PKG versions:
```
#if APT_PKG_MAJOR >= 5
#define pkgAcqFileSane pkgAcqFile
...

class pkgAcqFileSane : public pkgAcquire::Item
// This is frustrating: pkgAcqFile is **almost** good enough, but has some
// hardcoded stuff that makes it not quite work.
```
@mvo5 Is the workaround still needed at all, or could it be removed?

andy128k:
APT_PKG_MAJOR < 5 was 11 years ago.

mvo5:
Yeah, I'm kinda inclined to remove it and simplify things this way. I don't think we have user who would build against such an old apt (and we could always re-add it if someone comes along)

amaa-99:
@mvo5  I've removed that code and adjusted this pull request to:
`Remove the obsolete pkgAcqFile workaround implementation.`

mvo5 (approved):
Thanks!

PR #207, opened 2026-07-02
Revise the RPackage::installedFiles() implementation

amaa-99:
Changed the implementation to return std::string instead of relying on a static variable for storage lifetime.
Avoided calling FileExists() twice in case the file was already found.
Added returning an informative message in case the list can't be retrieved.
Added skipping the insertion of empty lines instead of relying on a single empty line always being read from the input file. Removed the unused HAVE_RPM variant of RPackage::installedFiles().
(closes amaa-99/synaptic#39)

### DIFF
--- a/common/rpackage.cc
+++ b/common/rpackage.cc
@@ -37,6 +37,7 @@
 #include <cstdio>
 #include <fstream>
 #include <map>
+#include <string>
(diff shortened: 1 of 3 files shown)


mvo5 (approved):
Nice!

PR #210, opened 2026-07-03
Remove using namespace from the header files

amaa-99:
Moved the 'using namespace std' statements out of the header files.
This keeps it from being applied as a blanket to every file which directly or indirectly includes one of those headers.
(closes amaa-99/synaptic#45)

### DIFF
--- a/common/rcdscanner.h
+++ b/common/rcdscanner.h
@@ -32,10 +32,7 @@
 
 #include "rpackagelister.h"
 
-using namespace std;
-
 class Configuration;
-
 
 class RCDScanProgress {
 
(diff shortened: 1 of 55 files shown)


common/rcdscanner.h
```diff
@@ -27,16 +27,13 @@
 
 #ifndef HAVE_APTPKG_CDROM
 
-#include <vector>
-#include <string>
+#include <std::vector>
```
mvo5: Hm, I don't think this will work (and the following line). I wonder why CI did not catch it. I guess the cdscanner is only used in the apt-rpm build, maybe worth adding that to CI too

amaa-99: Sorry, those are incorrect, going to fix it.

amaa-99: I see, I had those fixed a couple of commits later... Going to merge that into this pull.

amaa-99: Done.

mvo5 (approved):
Thanks!

PR #211, opened 2026-07-03
Set the gtk-disable-deprecated defines by default

amaa-99:
Changed the configuration so that the GTK deprecated functionality is disabled by default.
(closes amaa-99/synaptic#53)


mvo5 (approved):
This is a followup for one of the previous ones, right?

amaa-99:
> This is a followup for one of the previous ones, right?
Yes, I intended those to be enabled by default so we catch any issues early while developing.

PR #212, opened 2026-07-03
Add an option to put the application path definitions in the config.h (with default values)

amaa-99:
Added the --enable-ide-integration autogen option to enable putting the application path definitions (with default values) in the config.h file. This is useful for cases where the development tools fail due to those symbols not being defined.
(closes amaa-99/synaptic#56)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -172,6 +172,26 @@ AC_ARG_WITH(launchpad-integration,
 	]
 )
 
+# IDEs and code analisys tools need to know about the externally defined symbols,
+# so, for those scenarios, here those are added to the config.h file with default values.
+# Otherwise they're injected at build time (through the -D command line arguments in the makefiles).
+AC_ARG_ENABLE(ide-integration, [  --enable-ide-integration: Define default values in config.h for the injected symbols for the IDEs and code analisys tools.])
+AS_IF([test "x$enable_ide_integration" = "xyes"],
+    [
+        # These default values define relative paths to the current folder (i.e., where the executable is).
+        AC_DEFINE(PACKAGE_DATA_DIR, "./share", Default value for the IDEs and code analisys tools.)
+        AC_DEFINE(PACKAGE_LOCALE_DIR, "./share/locale", Default value for the IDEs and code analisys tools.)
+        AC_DEFINE(SYNAPTICSTATEDIR, "./var", Default value for the IDEs and code analisys tools.)
+        AC_DEFINE(SYNAPTIC_GTKBUILDERDIR, "./gtkbuilder/", Default value for the IDEs and code analisys tools.)
+        AC_DEFINE(SYNAPTIC_PIXMAPDIR, "./pixmaps/", Default value for the IDEs and code analisys tools.)
+    ],
+    [
+        # The paths defined here are replaced at make time by the standard values (as defined by GNU for the "Variables for Installation Directories").
+        # They can be adjusted by passing the --prefix option to autogen.
+        AC_SUBST(SYNAPTIC_APP_PATHS,'-DPACKAGE_DATA_DIR=\""${datadir}"\" -DPACKAGE_LOCALE_DIR=\""${localedir}"\" -DSYNAPTICSTATEDIR=\""${localstatedir}"\" -DSYNAPTIC_GTKBUILDERDIR=\""${datadir}/synaptic/gtkbuilder/"\" -DSYNAPTIC_PIXMAPDIR=\""${datadir}/synaptic/pixmaps/"\"')
+    ]
+)
+
 AC_ARG_WITH(pkg-hold,
         [  --with-pkg-hold: Build with experimental package "hold" feature],
         AC_DEFINE(SYNAPTIC_PKG_HOLD, 1, [build with package pin feature])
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -172,6 +172,26 @@ AC_ARG_WITH(launchpad-integration,
 	]
 )
 
+# IDEs and code analisys tools need to know about the externally defined symbols,
+# so, for those scenarios, here those are added to the config.h file with default values.
+# Otherwise they're injected at build time (through the -D command line arguments in the makefiles).
+AC_ARG_ENABLE(ide-integration,
```
mvo5: So I tried this with ./configure --disable-ide-integration and still have a `#define SYNAPTIC_GTKBUILDERDIR "./gtkbuilder/" ` in my config.h. Is there a `if test "x$enableval" = xyes; then` missing here? something like:
```
diff --git a/configure.ac b/configure.ac
index e68873bc..820a4043 100644
--- a/configure.ac
+++ b/configure.ac
@@ -178,12 +178,16 @@ AC_ARG_WITH(launchpad-integration,
 AC_ARG_ENABLE(ide-integration,
         [  --enable-ide-integration: Define default values in config.h for the injected symbols for the IDEs and code analisys tools.],
         [
-                # These default values define relative paths to the current folder (i.e., where the executable is).
-                AC_DEFINE(PACKAGE_DATA_DIR, "./share", Default value for the IDEs and code analisys tools.)
-                AC_DEFINE(PACKAGE_LOCALE_DIR, "./share/locale", Default value for the IDEs and code analisys tools.)
-                AC_DEFINE(SYNAPTICSTATEDIR, "./var", Default value for the IDEs and code analisys tools.)
-                AC_DEFINE(SYNAPTIC_GTKBUILDERDIR, "./gtkbuilder/", Default value for the IDEs and code analisys tools.)
-                AC_DEFINE(SYNAPTIC_PIXMAPDIR, "./pixmaps/", Default value for the IDEs and code analisys tools.)
+                if test "x$enableval" = xyes; then
+                        # These default values define relative paths to the current folder (i.e., where the executable is).
+                        AC_DEFINE(PACKAGE_DATA_DIR, "./share", Default value for the IDEs and code analisys tools.)
+                        AC_DEFINE(PACKAGE_LOCALE_DIR, "./share/locale", Default value for the IDEs and code analisys tools.)
+                        AC_DEFINE(SYNAPTICSTATEDIR, "./var", Default value for the IDEs and code analisys tools.)
+                        AC_DEFINE(SYNAPTIC_GTKBUILDERDIR, "./gtkbuilder/", Default value for the IDEs and code analisys tools.)
+                        AC_DEFINE(SYNAPTIC_PIXMAPDIR, "./pixmaps/", Default value for the IDEs and code analisys tools.)
+                else
+                        AC_SUBST(SYNAPTIC_APP_PATHS,'-DPACKAGE_DATA_DIR=\""${datadir}"\" -DPACKAGE_LOCALE_DIR=\""${localedir}"\" -DSYNAPTICSTATEDIR=\""${localstatedir}"\" -DSYNAPTIC_GTKBUILDERDIR=\""${datadir}/synaptic/gtkbuilder/"\" -DSYNAPTIC_PIXMAPDIR=\""${datadir}/synaptic/pixmaps/"\"')
+                fi
         ],
         [
                 # The paths defined here are replaced at make time by the standard values (as defined by GNU for the "Variables for Installation Directories").

```

mvo5: Actually I wonder if we should just focus on https://github.com/mvo5/synaptic/pull/186 its a bit more work to review but it seems nice to move to meson

amaa-99: I took the pattern from a couple of other similar options in the configure.ac where the checking for the `disable`/`without` is also missing (i.e. it expects only the 'enable' (or nothing) case).

Something in this direction would also be needed for the `meson` build, so I'd port it there too.
p.s. At the moment I'm still trying to workaround the reduced support for `meson` on the IDEs (apparently even less than for the `autotools`)...

amaa-99: I've changed that to the common textbook pattern (with the "AS_IF([test ...."). Both the `enable` and the `disable` should be working now (by default it's disabled).

amaa-99:
Until/if the project gets migrated to Meson (or something else) it would be great to have this in: It's the cleanest way that I found for having the project working well in any of the IDEs.

configure.ac
```diff
@@ -172,6 +172,26 @@ AC_ARG_WITH(launchpad-integration,
 	]
 )
 
+# IDEs and code analisys tools need to know about the externally defined symbols,
+# so, for those scenarios, here those are added to the config.h file with default values.
+# Otherwise they're injected at build time (through the -D command line arguments in the makefiles).
+AC_ARG_ENABLE(ide-integration, [  --enable-ide-integration: Define default values in config.h for the injected symbols for the IDEs and code analisys tools.])
+AS_IF([test "x$enable_ide_integration" == "xyes"],
```
mvo5: Tiny nitpick: on dash there is no `==` so this would break on ubuntu AFAICT. Just replacing it with `=` should be enough.

mvo5 (approved):

configure.ac
amaa-99: Done.

PR #216, opened 2026-07-06
Remove unused ui files

andy128k:


amaa-99:
@andy128k Those are obviously not referenced in the code :+1:, but, in case we'd want to check at build time, is there a way of verifying (at build time) that all needed files are there?

amaa-99 (approved):

mvo5 (approved):

andy128k:
@amaa-99 That's possible with `#embed` but it will be available in C++26 only. Well, it is also available in C23 but that's a different language.

PR #217, opened 2026-07-06
Fix the deprecated std::binary_function in the implementation of the std::sort custom comparator

amaa-99:
Replaced the implementation with one based on std::function.
(closes amaa-99/synaptic#91)

### DIFF
--- a/common/rpackagelister.cc
+++ b/common/rpackagelister.cc
@@ -932,14 +932,6 @@ void RPackageLister::getSummary(int &held, int &kept, int &essential,
    sizeChange = deps->UsrSize();
 }
 
-
-
-struct bla:public binary_function<RPackage *, RPackage *, bool> {
-   bool operator() (RPackage *a, RPackage *b) {
-      return strcmp(a->name(), b->name()) < 0;
-   }
-};
-
 void RPackageLister::saveUndoState(pkgState &state)
 {
    undoStack.push_front(state);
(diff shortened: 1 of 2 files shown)


amaa-99:
p.s. I would have preferred using the std comparators instead, but those operate on std::string parameters (the RPackage name() returns char * instead).

common/rpackagelister.cc
```diff
@@ -932,12 +932,10 @@ void RPackageLister::getSummary(int &held, int &kept, int &essential,
...
 }
 
-
-
-struct bla:public binary_function<RPackage *, RPackage *, bool> {
-   bool operator() (RPackage *a, RPackage *b) {
-      return strcmp(a->name(), b->name()) < 0;
-   }
+// Custom sort functor to be used with std::sort.
+std::function<bool(RPackage *a, RPackage *b)> RPackageLister::bla = [](RPackage *a, RPackage *b)
```
andy128k: No need to use `std::function` either. This is not a closure, so a regular function could be used instead.

amaa-99: That's what I thought, but I got a weird compilation error... Does it work on your env with a regular function?

amaa-99: Meh, found the error... I had const on the params...
(That will be another fix later on.)

amaa-99: I'll have a look at the missing const attributes of the getters sometime soon...

common/rpackagelister.h
```diff
@@ -272,7 +272,8 @@ class RPackageLister {
                            double &sizeChange);
 
    void getDownloadSummary(int &dlCount, double &dlSize);
-
+   
+   static bool bla(RPackage *a, RPackage *b);
```
andy128k: It does not need to be a member.

common/rpackagelister.cc
```diff
@@ -932,12 +932,10 @@ void RPackageLister::getSummary(int &held, int &kept, int &essential,
...
 }
 
-
-
-struct bla:public binary_function<RPackage *, RPackage *, bool> {
-   bool operator() (RPackage *a, RPackage *b) {
-      return strcmp(a->name(), b->name()) < 0;
-   }
+// Custom sort functor to be used with std::sort.
+bool RPackageLister::bla(RPackage *a, RPackage *b)
```
andy128k: May I suggest a better name? How about `byNameAscending`?

amaa-99: Why not, since we're already touching that...

common/rpackagelister.h
amaa-99: I tend to prefer having everything scoped to the respective class (i.e. as a member), but for private static members it's good either way.
If you don´t have a big pref I'd keep as a member...

mvo5:
This looks fine to me, if @andy128k is also happy I will merge :)

common/rpackagelister.h
andy128k: Well, it needlessly pollutes an interface and exposes an implementation detail. If you want it to be scoped than I'd suggest to move it to `RPackage` class and make it public.

amaa-99: The exposition of private members on c++ is something I've never been happy with but I've made my peace with it long ago ;). In this case RPackage is a fitting place for this method :+1: - Going to move it there.

mvo5 (approved):
Thank you!

PR #218, opened 2026-07-06
Replace message box about not beeing root with an infobar

andy128k:
image

### DIFF
--- a/gtk/gsynaptic.cc
+++ b/gtk/gsynaptic.cc
@@ -446,18 +446,6 @@ static gint applicationHandleLocalOptions(GApplication* app,
 
 static void applicationStartup(GApplication* app, gpointer user_data)
 {
-   if (getuid() != 0) {
-      RGUserDialog userDialog;
-      userDialog.warning(g_strdup_printf("<b><big>%s</big></b>\n\n%s",
-                                         _("Starting \"Synaptic Package Manager\" without "
-                                           "administrative privileges"),
-				         _("You will not be able to apply "
-				           "any changes, but you can still "
-					   "export the marked changes or "
-					   "create a download script "
-					   "for them.")));
-   }
-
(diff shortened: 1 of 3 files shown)


amaa-99:
image

The style of the text is a bit off on my kde plasma env.: Orange text while the window is selected.
@andy128k Do you know where is that styling coming from?

amaa-99:
The Gtk.InfoBar is marked as deprecated on 4.1 (https://docs.gtk.org/gtk4/class.InfoBar.html).

amaa-99:
I'd suggest the following text instead:
Running without administrative privileges. Changes can't be applied, but marked changes can still be exported and download scripts can be created for them.

amaa-99:
What do you think of instead showing this on the status bar (at the bottom)?

For example the "Running without administrative privileges" (plus eventually a yellow warning icon), and popping up on hover the "Changes can't be applied, but marked changes can still be exported and a download script can be created for them." message?

image

image

amaa-99:
p.s. Not having the popup/dialog on startup is definitely nicer :+1:!

mvo5:
I like this UI, it looks very nice! A shame that they deprecated the InfoBar :( What is there replacement? Or there is none?

andy128k:
Gtk+3 shows its age. It suppose to have a blue background but fell back to a peach on my machine.

@mvo5 The replacement would be a regular Box with custom CSS or [Adwaita's Toast and Banner.](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.9/widget-gallery.html#toast-overlay).

andy128k:
I've replaced InfoBar with a regular Box and custom CSS. Now it should look the same everywhere.

amaa-99:
Eventually adding the "not running with administrator privileges" to the (main) window title too (in any of the variants)?

PR #219, opened 2026-07-06
Install synaptic into /bin instead of /sbin

andy128k:
Install synaptic into `/bin` instead of `/sbin`.

There are a few reasons to do so:

1. The program is already works (in a limited way) for non-privileged users.

An excerpt from FHS:
> Deciding what things go into "sbin" directories is simple: if a normal (not a system administrator) user will ever run it directly, then it must be placed in one of the "bin" directories. Ordinary users should not have to place any of the sbin directories in their path.
> For example, files such as chfn which users only occasionally use must still be placed in /usr/bin. ping, although it is absolutely necessary for root (network recovery and diagnosis) is often used by users and must live in /bin for that reason

2. In the future running GUI application with root privileges will be problematic. Wayland already forbids this and relying on XWayland is rather a workaround,

### DIFF
--- a/data/com.ubuntu.pkexec.synaptic.policy.in
+++ b/data/com.ubuntu.pkexec.synaptic.policy.in
@@ -12,7 +12,7 @@
       <allow_inactive>auth_admin</allow_inactive>
       <allow_active>auth_admin</allow_active>
     </defaults>
-    <annotate key="org.freedesktop.policykit.exec.path">/usr/sbin/synaptic</annotate>
+    <annotate key="org.freedesktop.policykit.exec.path">/usr/bin/synaptic</annotate>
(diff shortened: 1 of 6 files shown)


mvo5 (approved):
Thanks, I think this makes sense!
 
And yes, I'm aware of the long term implications with root and wayland :( Sadly its a bit of work to fix it, we would have to add a varlink/dbus backend that does all the priviliged stuff and have the gui as non-priv. However our gtklist would probably be performing extremly poorly in this case when it displays all >60k packages - but maybe worth a shot? idk, nowdays I would go with varlink as the IPC for its simplicity (and speed).

amaa-99:
@andy128k In an update scenario, will the file stored in the sbin get removed?

andy128k:
@amaa-99 That's what packagers should worry about :smile:

PR #220, opened 2026-07-07
Evaluate a sqlite based version of the quick filter (feedback wanted)

amaa-99:
Test implementation of the quick filter using sqlite and fts5.

I've picked on the idea from @julian-klode and put up a version for evaluating the usability.
Could you (@mvo5, @andy128k) give this a try and let me know if I should refine this further?

Sample syntax (from the fts5 specs):
`"linux-"` - All packages containing `linux-` in the name or description
`^"linux-"` -  All packages starting with `linux-` in the name or description
`name:^"linux-"` - All packages starting with `linux-` in the name
`description:^"linux-"` - All packages starting with `linux-` in the description

### DIFF
--- a/common/i18n.h
+++ b/common/i18n.h
@@ -24,7 +24,7 @@
 #define _I18N_H_
 
 #ifdef HAVE_CONFIG_H
-#include "config.h"
+#include "config.h"  // IWYU pragma: associated
 #endif
 
 #ifdef ENABLE_NLS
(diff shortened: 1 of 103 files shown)


amaa-99:
The populating (and indexing) of the text search database is done synchronously whenever the cache gets updated (at the same spot where the view gets updated). On my tests it took up to 2s to populate/index the database, so the extra delay is not particularly annoying.

If this is a viable option I'd look into running the indexing asynchronously, and having a parser for the search string.

amaa-99:
Dropping this for now (too many conflicts, needs re-implementation).

PR #221, opened 2026-07-08
Replace message box about not beeing root with a message in status bar

andy128k:
An alternative to #218

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -18,7 +18,7 @@ jobs:
         sudo apt update
         sudo apt build-dep -y ./
         #Devscripts for bringing in debuild
-        sudo apt install meson ninja-build appstream libpolkit-gobject-1-dev devscripts -y
+        sudo apt install meson ninja-build appstream desktop-file-utils libpolkit-gobject-1-dev devscripts -y
 
     - name: configure
       run: |
--- a/gtk/gsynaptic.cc
+++ b/gtk/gsynaptic.cc
@@ -397,17 +397,7 @@ static gint applicationHandleLocalOptions(GApplication* app,
 
 static void applicationStartup(GApplication* app, gpointer user_data)
 {
-   if (getuid() != 0) {
-      RGUserDialog userDialog;
-      userDialog.warning(g_strdup_printf("<b><big>%s</big></b>\n\n%s",
-                                         _("Starting \"Synaptic Package Manager\" without "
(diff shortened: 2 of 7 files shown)


mvo5 (approved):
I like it - I wonder if we should make it bold or somehow more visible - that was the advantage of the other approach, it was more "in your face" :) Maybe with a little bit of tweaking we can archive the same. But the direction is nice and it is indeed a bit silly to always show this dialog.

andy128k:
@mvo5  It's up to you which one to pick. The other one works already and this one has issues with autotools (a similar change on top of meson works well locally).

amaa-99:
I prefer it on the status bar: Still pretty visible, but not taking additional space. ;)
Eventually with adding the "not running with administrator privileges" to the (main) window title too (in any of the variants).

amaa-99:
Looking like this on my env:
image
@andy128k As intended?

mvo5:
@andy128k Thank you! I have no super strong opinion, I will go with this one as the other one is deprecated and its probably visible enough (we will get bugreports if not ;) - so if this looks good and if the screenshot of @amaa-99 (thanks for testing this!) looks correct I'm happy to merge.

Thank you both for bringing so much life back into the project, really appreciated!

amaa-99:
@andy128k The warning icon is missing on the screenshot: The `gtk-update-icon-cache` fails in case the folder tree doesn't already exist (i.e. the `/share/synaptic/icons/hicolor`): `..../share/synaptic/icons/hicolor/.icon-theme.cache : No such file or directory`
Creating the folder in advance (in case it doesn't exist) should fix that.
p.s. I'm using a custom install 'prefix' (i.e. `autogen.sh --prefix .....`).

This seems to do the trick (in the `.am` file):
```
install-data-local:
	@if test -z "$(DESTDIR)"; then $(MKDIR_P) $(datadir)/synaptic/icons/hicolor && $(GTK_UPDATE_ICON_CACHE) -q -t -f $(datadir)/synaptic/icons/hicolor ; fi
```

amaa-99:
p.s. There's another 'warning' icon already in use:
image
Should we use the same one for both places?

andy128k:
@amaa-99  It's a stock image (not part of the distribution). And stock images are deprecated since Gtk+ 3.10 (2013). So, rework of application's assets is needed anyway.

mvo5 (approved):
Thanks! Still a bit on the fence which version I like better but given that this is ready and has a +1 already lets start here.

PR #222, opened 2026-07-09
Replace the deprecated GdkColor

amaa-99:
Replaced GdkColor with GdkRGBA.
Moved the initialization of _gray to the declaration and declared it const.

(closes amaa-99/synaptic#95)


mvo5 (approved):
Nice

PR #223, opened 2026-07-09
Revise the #include statements

amaa-99:
Added missing #include statements and removed the not used ones.
Grouped the #include statements and removed duplicates.
Fixed the circular/recursive inclusion cases.

(closes amaa-99/synaptic#97)

### DIFF
--- a/common/rinstallprogress.cc
+++ b/common/rinstallprogress.cc
@@ -22,26 +22,30 @@
  * USA
  */
 
-#include "config.h"
-
-#include <unistd.h>
-#include <fcntl.h>
-#include <sys/types.h>
-#include <sys/wait.h>
-#include <iostream>
-#include <cstdio>
-#include <apt-pkg/error.h>
-#include <apt-pkg/install-progress.h>
-#ifdef HAVE_RPM
-#include <apt-pkg/configuration.h>
-#endif
+#include "config.h"  // IWYU pragma: associated
 
 #include "rinstallprogress.h"
 
 #include "i18n.h"
-std::string RInstallProgress::finishMsg = _("\nSuccessfully applied all changes. You can close the window now.");
-std::string RInstallProgress::errorMsg = _("\nNot all changes and updates succeeded. For further details of the failure, please expand the 'Details' panel below.");
-std::string RInstallProgress::incompleteMsg = 
+
+#include <apt-pkg/install-progress.h>
+#include <apt-pkg/packagemanager.h>
+#include <stdlib.h>
+#include <string>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#ifdef HAVE_RPM
+#include <apt-pkg/configuration.h>
+#include <fcntl.h>
+#endif
+
+using namespace std;
+
+string RInstallProgress::finishMsg = _("\nSuccessfully applied all changes. You can close the window now.");
+string RInstallProgress::errorMsg = _("\nNot all changes and updates succeeded. For further details of the failure, please expand the 'Details' panel below.");
+string RInstallProgress::incompleteMsg = 
       _("\nSuccessfully installed all packages of the current medium. "
 	"To continue the installation with the next medium close "
 	"this window.");
@@ -63,7 +67,6 @@ const char* RInstallProgress::getResultStr(pkgPackageManager::OrderResult res)
 
    return "Unknown install result.";
 }
-
 
 pkgPackageManager::OrderResult RInstallProgress::start(pkgPackageManager *pm,
                                                        int numPackages,
--- a/gtk/rgcdscanner.h
+++ b/gtk/rgcdscanner.h
@@ -24,12 +24,19 @@
 #ifndef RGCDSCANNER_H
 #define RGCDSCANNER_H
 
-#include "config.h"
+#include "config.h"  // IWYU pragma: associated
 
 #ifndef HAVE_APTPKG_CDROM
 
 #include "rcdscanner.h"
-#include "rggtkbuilderwindow.h"
+
+#include "rgwindow.h"
+#include "ruserdialog.h"
+
+#include <gtk/gtk.h>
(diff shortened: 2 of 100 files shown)


common/rinstallprogress.cc
```diff
@@ -22,26 +22,30 @@
...
-#include "rinstallprogress.h"
+using namespace std;
 
-#include "i18n.h"
-std::string RInstallProgress::finishMsg = _("\nSuccessfully applied all changes. You can close the window now.");
-std::string RInstallProgress::errorMsg = _("\nNot all changes and updates succeeded. For further details of the failure, please expand the 'Details' panel below.");
-std::string RInstallProgress::incompleteMsg = 
+string RInstallProgress::finishMsg = _("\nSuccessfully applied all changes. You can close the window now.");
+string RInstallProgress::errorMsg = _("\nNot all changes and updates succeeded. For further details of the failure, please expand the 'Details' panel below.");
+string RInstallProgress::incompleteMsg =
```
andy128k, line 48: This seems like a step back. Namespace qualifiers are already enforced in headers.  Why to lower the bar for `.cc` files?

gtk/rgpreferenceswindow.cc
```diff
@@ -22,22 +22,34 @@
...
+#include <apt-pkg/error.h>
+#include <apt-pkg/strutl.h>
+#include <cassert>
+#include <cstdlib>
+#include <cstring>
+#include <gdk/gdk.h>
+#include <glib-object.h>
+#include <glib.h>
+#include <glib/gtypes.h>
+#include <gobject/gclosure.h>
```
andy128k, line 48: This one is not needed.. When it comes to GLib and Gtk, just two includes are needed (<glib.h> and <gtk/gtk.h>).

andy128k (commented):
+1075 -779 :fearful:

A lot of added includes are redundant.  E,g, regular expressions are used in 3 modules but there are 6 includes,

common/rinstallprogress.cc
amaa-99: The 'using namespace' statements in the headers files should be avoided since they propagate as a blanket to anything including that file (directly or indirectly). In the other hand, in the .c files the 'using namespace' is contained so it's more acceptable.

In this cleanup I kept the pattern already used in most .c files, with the `using namespace std`.
I prefer in general using the (full) qualifiers as much as it's practical, so for things like std:: I'd go for the full version.

I'd consider changing that pattern for all (.c) files, but in a separate pull request (for the std:: members). What do you think?

gtk/rgpreferenceswindow.cc
amaa-99: I tried to get a balance between keeping the static analysis tools happy and going down the rabbit :rabbit: hole.
`glib-object.h` is not directly included by glib.h, so I get why the static analysis would not consider it iwyu in general, but for glib/gtypes.h the tools should be able to identify that that's an aggregate include file.
The gtk.h sub-includes seem to be accepted fine if one includes only the gtk.h though.

amaa-99: I'm not happy with those extra `glib/gtypes.h` in there either. Should those get refined "by hand" and we live with the extra warnings from the analysis tools? Or have the redundant ones and be able to see when there's an important warning? What's your opinion?

amaa-99:
> +1075 -779 😨
> 
> A lot of added includes are redundant. E,g, regular expressions are used in 3 modules but there are 6 includes,

I had a look at the <regex.h> includes: They all seem to match used references in the respective files...
Could you point me to one of the unused ones? Maybe I'm looking at a different include?

amaa-99:
@mvo5 Hi Michael. I know it's a somewhat larger pull but it would be great if we could get this in soon.
With these changes I got the IDEs and code analysis tools working pretty reliably.
At the moment I have to keep jumping between worktrees with theses changes and without while going after code issues.(Also I'd like to avoid having to re-revise all of this if we wait too long...)

gtk/rgcdscanner.h
```diff
@@ -24,12 +24,19 @@
...
 #include "rcdscanner.h"
-#include "rggtkbuilderwindow.h"
+
+#include "rgwindow.h"
+#include "ruserdialog.h"
+
+#include <gtk/gtk.h>
+#include <string>
+
+using namespace std;
```
mvo5, line 39: Is this an oversight? This PR ususally did it the other way around (the using namespace std into the .cc)?

gtk/rguserdialog.cc
```diff
@@ -22,20 +22,27 @@
...
+#include <gdk-pixbuf/gdk-pixbuf.h>
+#include <gdk/gdk.h>
 #include <gdk/gdkx.h>
-
-#include <assert.h>
+#include <glib-object.h>
+#include <glib.h>
+#include <gobject/gclosure.h>
+#include <gtk/gtk.h>
+#include <ruserdialog.h>
```
mvo5, line 44: This is duplicated, right? we have "#include "rguserdialog.h"" a few lines above?

mvo5 (approved):
Thanks, I will merge this to unblock you but here are some small tweaks needed but please just do the post-merge as a followup (see the inline comments). It would be nice if future PRs like this would avoid mixing the header changes with extra whitespace changes (maybe your editor did that, no idea). Its just easier to review this way :) 

Another idea that I would really appreciated: now that we have include-what-you-need clean code, can we add a github worklflow so that we run the same check you ran locally on every PR run to ensure we do not regress here? Same for the other static analysis tool(s) you are using? 

Thanks for the cleanups, much appreciate the work!

gtk/rguserdialog.cc
amaa-99: ;) `rguserdialog.h` and `ruserdialog.h`, both needed.
The `ruserdialog.h` isn't a system header though - already fixed it.

amaa-99:
Many thanks!

PR #225, opened 2026-07-09
Fix the "enumeration values 'ButtonsDefault' and 'ButtonsOk' not handled in switch"

amaa-99:
Added the explicit default to the switch statement.
(closes amaa-99/synaptic#106)

### DIFF
--- a/gtk/rguserdialog.cc
+++ b/gtk/rguserdialog.cc
@@ -130,7 +130,7 @@ bool RGUserDialog::showErrors()
 
 bool RGUserDialog::message(const char *msg,
                            RUserDialog::DialogType dialog,
-                           RUserDialog::ButtonsType buttons, bool defres)
+                           RUserDialog::ButtonsType buttons, bool defaultResponse)
 {
    GtkWidget *dia;
    GtkResponseType res;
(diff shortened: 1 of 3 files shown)


gtk/rguserdialog.h
```diff
@@ -47,7 +47,7 @@ class RGUserDialog : public RUserDialog
     virtual bool message(const char *msg,
 	    RUserDialog::DialogType dialog=RUserDialog::DialogInfo,
 	    RUserDialog::ButtonsType buttons=RUserDialog::ButtonsOk,
-	    bool defres=true);
+	    bool defautResponse=true);
```
mvo5: nitpick: typo defaut -> default (missing "l")

gtk/rguserdialog.cc
```diff
@@ -170,33 +170,26 @@ bool RGUserDialog::message(const char *msg,
...
-   if (defres) {
-      switch (buttons) {
-         case RUserDialog::ButtonsOkCancel:
-            gtk_dialog_set_default_response(GTK_DIALOG(dia), GTK_RESPONSE_OK);
-            break;
-         case RUserDialog::ButtonsYesNo:
-            gtk_dialog_set_default_response(GTK_DIALOG(dia), GTK_RESPONSE_YES);
-            break;
+   switch (buttons) {
+      case RUserDialog::ButtonsOkCancel: {
```
mvo5, line 183: (nitpick): the `{` is not needed here (and uncommon in C++ IME). This:
```C++
   switch (buttons) {
      case RUserDialog::ButtonsOkCancel:
         gtk_dialog_set_default_response(GTK_DIALOG(dia), defaultResponse ? GTK_RESPONSE_OK : GTK_RESPONSE_CANCEL);
         break;
      case RUserDialog::ButtonsYesNo:
         gtk_dialog_set_default_response(GTK_DIALOG(dia), defaultResponse ? GTK_RESPONSE_YES : GTK_RESPONSE_NO);
         break;
      default:
         break;
   }
```
should be fine :)

mvo5 (commented):
Thanks! This is nice. While work on this, maybe worth renaming "common/ruserdialog.h" defres->defaultReponse there too. Two small inline suggesitons.

gtk/rguserdialog.h
amaa-99: I thought I fixed that. Thanks for noticing!

gtk/rguserdialog.cc
amaa-99: It's for the humans (reading the code) ;).
Adding braces to `case`, `if` and similar cases helps preventing some weird bugs.
Also many IDEs/editors use the braces for the code folding and tracking code blocks.
In some scenarios the scoping helps the compiler, but not here.

amaa-99: p.s. Many thanks for noticing this kind of stuff and commenting :+1: I love the feedback!

gtk/rguserdialog.h
amaa-99: Done!

mvo5 (approved):
thanks!

PR #226, opened 2026-07-09
Fix "object backing the pointer will be destroyed at the end of the full-expression"

amaa-99:
Fixed the access to the object's data outside its lifetime.
(closes amaa-99/synaptic#103)


amaa-99:
Another one ;).

mvo5 (approved):
Thanks!

PR #227, opened 2026-07-09
Fix "unused local variable"

amaa-99:
Removed the unused (duplicate) variables.
Adjusted the variable names to lower case and moved the local string variable into the loop.
Refactored the (redundant) comparisons against true.

(closes amaa-99/synaptic#109)

### DIFF
--- a/common/rcacheactor.cc
+++ b/common/rcacheactor.cc
@@ -52,50 +52,48 @@ void RCacheActor::notifyCachePostChange()
 
 
 RCacheActorRecommends::RCacheActorRecommends(RPackageLister *lister,
-                                             string FileName)
+                                             string fileName)
 : RCacheActor(lister)
 {
-   FileFd F(FileName, FileFd::ReadOnly);
+   FileFd f(fileName, FileFd::ReadOnly);
    if (_error->PendingError()) {
-      _error->Error("could not open recommends file %s", FileName.c_str());
+      _error->Error("Could not open recommends file %s", fileName.c_str());
       return;
    }
-   pkgTagFile Tags(&F);
-   pkgTagSection Section;
+   
+   pkgTagFile tags(&f);
+   pkgTagSection section;
 
-   string Name;
-   string Match;
-   string Recommends;
-   while (Tags.Step(Section)) {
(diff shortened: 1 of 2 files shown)


common/rcacheactor.cc
```diff
@@ -52,50 +52,48 @@ void RCacheActor::notifyCachePostChange()
...
-      if (Name.empty() == true || Recommends.empty() == true)
-         continue;
-
-      ListType *List = NULL;
-      if (Match.empty() == true || Match == "exact") {
-         List = &_map[Name];
-      } else if (Match == "wildcard") {
-         List = &_map_wildcard[Name];
-      } else if (Match == "regex") {
+
```
mvo5, line 63: super nitpick: whitespace here

common/rcacheactor.cc
```diff
@@ -52,50 +52,48 @@ void RCacheActor::notifyCachePostChange()
 
 
 RCacheActorRecommends::RCacheActorRecommends(RPackageLister *lister,
-                                             string FileName)
+                                             string fileName)
 : RCacheActor(lister)
 {
-   FileFd F(FileName, FileFd::ReadOnly);
+   FileFd f(fileName, FileFd::ReadOnly);
```
mvo5, line 58: I'm not 100% sure the renames are worth it (FileName -> filename, F -> f) because the diff is slightly harder to read this way. All looks correct so I'm inclined to just take it as is (to avoid that you have extra work). But maybe just apply this diff? 

[preserve-names2.diff.txt](https://github.com/user-attachments/files/29856980/preserve-names2.diff.txt) it would make the review easier.

mvo5 (commented):
Thanks! Some inline suggestions.

common/rcacheactor.cc
amaa-99: In general I would have left it, but I suspect that the Uppercase variables where part of why no one noticed the issue(s) there.

common/rcacheactor.cc
amaa-99: Just a small separator for the next logical unit.

mvo5 (approved):
Thanks!

PR #228, opened 2026-07-10
Fix "unused local variable" strVal in setTreeList()

amaa-99:
Removed the unused variable strVal in setTreeList() and some cleanups in the surrounding code.

(closes amaa-99/synaptic#115)

PR #229, opened 2026-07-10
Fix "unused local variable" VerTag in getAvailableVersions()

amaa-99:
Removed the unused variables VerTag in getAvailableVersions.
(closes amaa-99/synaptic#112)


mvo5:
This looks identical to #228 - am I missing something subtle here?

amaa-99:
@mvo5 Sorry, added the same commit on both pulls. Already corrected it!

PR #231, opened 2026-07-10
Fix "unused local variable" OrigPath in rpmindexcopy.cc

amaa-99:
Removed the unused variables.
(closes amaa-99/synaptic#120)


mvo5 (approved):
Thanks!

PR #232, opened 2026-07-10
Fix "unused local variable" in rpackagefilter.cc

amaa-99:
Removed the unused variables and corrected int to size_t.
(closes amaa-99/synaptic#122)

### DIFF
--- a/common/rpackagefilter.cc
+++ b/common/rpackagefilter.cc
@@ -452,18 +452,15 @@ void RPatternPackageFilter::addPattern(DepType type, string pattern,
 
 bool RPatternPackageFilter::write(ofstream &out, string pad)
 {
-   DepType type;
-   string pat;
-   bool excl;
-
    out << pad + "andMode " << and_mode << ";" << endl;
-
    out << pad + "patterns {" << endl;
 
    for (size_t i = 0; i < count(); i++) {
+      DepType type;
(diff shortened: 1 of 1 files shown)


mvo5 (commented):
Thanks, for this one it really would be better to split into two PRs or two commits. Ideally the removal of the unused var would be its own commit and the change to format its own as well. But I don't want to impose more work on you so fine to keep things a bit muddled together. But here its the int->size_t and then the unused var, that is a bit much :) Sorry for the extra work this causes.

amaa-99:
> Thanks, for this one it really would be better to split into two PRs or two commits. Ideally the removal of the unused var would be its own commit and the change to format its own as well. But I don't want to impose more work on you so fine to keep things a bit muddled together. But here its the int->size_t and then the unused var, that is a bit much :) Sorry for the extra work this causes.

No worries. In general I also prefer separating... But in cases like this there's the issue of the merge conflict that will likely happen (and I also try to avoid giving you the effort of revising the same code twice in a row)...
Let's split it and see how it goes...

p.s. Normally easily done... At the moment I'm kind of IDEless ;).

amaa-99:
@mvo5 There's a new pull for the rest (#248).

mvo5 (approved):
Thank you

PR #233, opened 2026-07-10
Fix unitialized local variable "type" in RGFilterManagerWindow::getPatternFilter

amaa-99:
Fixed the potential cases where 'type' would be used unitialized or with the wrong value.
Removed the unused local variables.
(closes amaa-99/synaptic#124)

### DIFF
--- a/gtk/rgfiltermanager.cc
+++ b/gtk/rgfiltermanager.cc
@@ -718,53 +718,42 @@ void RGFilterManagerWindow::getStatusFilter(RStatusPackageFilter & f)
 
 void RGFilterManagerWindow::getPatternFilter(RPatternPackageFilter &f)
 {
-   GtkTreeIter iter;
-   bool exclude;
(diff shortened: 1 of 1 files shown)


gtk/rgfiltermanager.cc
```diff
@@ -719,53 +719,42 @@ void RGFilterManagerWindow::getStatusFilter(RStatusPackageFilter & f)
 
 void RGFilterManagerWindow::getPatternFilter(RPatternPackageFilter &f)
 {
-   GtkTreeIter iter;
-   bool exclude;
-   string pattern;
-   RPatternPackageFilter::DepType type;
-   gchar *dopatt, *what, *text;
```
amaa-99, line 725: In case there's no match type will be used uninitialized or with the previous (wrong) value.

mvo5 (approved):
Thank you!

PR #235, opened 2026-07-12
Fix empty element in the sections list

amaa-99:
In case there are packages with empty sections field the sections list shows an empty row which misbehaves:
\
image
\
Additionally, the sorting doesn't handle well differences on the case (upper/lower) of the first letter on the section name:
\
image
\
Modified the implementation so that empty section strings default to 'Unknown'.
Modified the implementation to capitalize the first letter of the displayed section strings so that the sorting is consistent.
Minor refactoring for performance and code safety.
(closes amaa-99/synaptic#126)

### DIFF
--- a/common/rpackageview.cc
+++ b/common/rpackageview.cc
@@ -30,7 +30,6 @@
 #include "rconfiguration.h"
 #include "rpackage.h"
 #include "rpackagefilter.h"
-#include "sections_trans.h"
 
 #include <algorithm>
 #include <apt-pkg/configuration.h>
@@ -103,12 +102,6 @@ void RPackageView::refresh()
       if (_all[i])
          addPackage(_all[i]);
    }
-}
-
-void RPackageViewSections::addPackage(RPackage *package)
-{
-   string str = trans_section(package->section());
-   _view[str].push_back(package);
 }
 
 RPackageViewStatus::RPackageViewStatus(vector<RPackage *> &allPkgs)
--- a/common/rpackageview.h
+++ b/common/rpackageview.h
@@ -28,6 +28,7 @@
 
 #include "i18n.h"
 #include "rpackage.h"
+#include "sections_trans.h"
 
 #include <cctype>
 #include <cstddef>
@@ -109,7 +110,6 @@ class RPackageView
    virtual void refresh();
 };
 
-
 class RPackageViewSections : public RPackageView
 {
  public:
@@ -122,7 +122,12 @@ class RPackageViewSections : public RPackageView
       return _("Sections");
    };
 
-   void addPackage(RPackage *package);
+   inline void addPackage(RPackage *package)
+   {
+      std::string section = trans_section(package->section());
+      section[0] = static_cast<char>(std::toupper(static_cast<unsigned char>(section[0])));
+      _view[section].push_back(package);
+   }
 };
 
 class RPackageViewAlphabetic : public RPackageView
--- a/common/sections_trans.cc
+++ b/common/sections_trans.cc
@@ -9,10 +9,9 @@
 
 #include "i18n.h"
 
-#include <apt-pkg/strutl.h>
 #include <cstddef>
-#include <sstream>
 #include <string>
+#include <string_view>
 
 using namespace std;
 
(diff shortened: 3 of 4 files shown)


amaa-99:
I'm practically on `vi` mode at the moment (haven't fixed the IDEs to work with meson yet), but I'll try to fix those whitespaces caught by the lint 👍.
@mvo5 Thanks for adding the check to the CI, it's one thing less to track manually.

common/rpackageview.h
```diff
@@ -104,7 +103,12 @@ class RPackageViewSections : public RPackageView {
       return _("Sections");
    };
 
-   void addPackage(RPackage *package);
+   inline void addPackage(RPackage *package)
+   {
+      std::string section = trans_section(package->section());
+      section[0] = static_cast<char>(std::toupper(static_cast<unsigned char>(section[0])));
```
andy128k, line 128: This will corrupt a message in a non-latin language.

andy128k (changes requested):

andy128k:
Since "Unknown" is not present in a table of translated messages, it should be translated separately.

common/rpackageview.h
andy128k: Generally it's a bad idea to manipulate a translatable text. You cannot guess what language it is and how words are structured there. They may not have any notion of "capitalization" at all. Or be written in a different direction. Or have different order of words. Basically nothing could be guessed and should be left to translators.

amaa-99: Per specification std::toupper will not touch any of those characters (which is the biggest weakness since it won't capitalize an ü for example).
The correction is done on the translated text, so it won´t mess any lookups for the translation.

Update:

> Converts the given character to uppercase according to the character conversion rules defined by the currently installed C locale.
> 
> In the default "C" locale, the following lowercase letters abcdefghijklmnopqrstuvwxyz are replaced with respective uppercase letters ABCDEFGHIJKLMNOPQRSTUVWXYZ.

amaa-99:
> Since "Unknown" is not present in a table of translated messages, it should be translated separately.

I took 'Unknown' as a default since there are already some packages using that. Considering those names come from external sources I guess there will always be some that have no translation...

common/rpackageview.h
andy128k: Modern systems use UTF-8 where a character may occupy more than one byte, so technically this code modifiers part of the character. This code will not break UTF-8 string, but I still believe and insist that capitalization (on any other similar transformation) should be done in translations and not in the code.

andy128k (changes requested):

common/sections_trans.cc
```diff
@@ -151,39 +150,35 @@ const char *transtable[][2] = {
...
+   for (size_t i = 0; transtable[i][0] != NULL; i++) {
+      if (s == transtable[i][0]) return _(transtable[i][1]);
+   }
+
+   return s;   // No translation found: Return the original string.
+}
+
+string trans_section(const string &section)
+{
+   string_view str = (!section.empty()) ? string_view(section) : string_view("Unknown");
```
andy128k, line 166:
```suggestion
   if (section.empty()) return _("Unknown");
   string_view str = string_view(section);
```

amaa-99: The other cases of "Unknown" would go through the other (translation) path...
Wouldn't it be better to have both translated in the 'same way'?

amaa-99: I'm inclined to putting those in the
> [Miscellaneous](https://packages.debian.org/unstable/misc/)
>     Miscellaneous utilities that didn't fit well anywhere else.

section. That way they'd be in one of the "official/existing" sections.
@andy128k @mvo5 What do you think?

common/rpackageview.cc
```diff
@@ -66,10 +65,10 @@ bool RPackageView::setSelected(string name)
 
 vector<string> RPackageView::getSubViews()
 {
-   vector<string> subViews;
-   for (map<string, vector<RPackage *> >::iterator I = _view.begin();
-        I != _view.end(); I++)
-      subViews.push_back((*I).first);
+   vector<string> subViews(_view.size());
+   for (const std::pair<string, vector<RPackage *>> &subView : _view) {
```
mvo5, line 70: I guess we could nowdays also:
```
  for (const auto &subView : _view) {
```
here?

amaa-99: In general, explicit types are preferred wherever practical, auto in the rare cases where it must be...
When refactoring code, changing autos to explicit types would be one of the tasks ;).

common/sections_trans.cc
amaa-99: @mvo5 Any thoughts about the translation/default section name here?

amaa-99:
Dropping this for now. Will re-implement after I'm back (if the issue is still present).

PR #236, opened 2026-07-13
meson: tweak to make the deb build identical to before

mvo5:
With this branch:
```
$ debdiff /var/cache/pbuilder/result/synaptic_0.91.6_amd64.deb /var/cache/pbuilder/result/synaptic_0.91.7_amd64.deb
[The following lists of changes regard files as different if they have
different names, permissions or owners.]

Files in second .deb but not in first
-------------------------------------
-rw-r--r--  root/root   /usr/share/synaptic/html/synaptic.proc
-rwxr-xr-x  root/root   /usr/bin/synaptic

Files in first .deb but not in second
-------------------------------------
-rw-r--r--  root/root   /usr/share/synaptic/gtkbuilder/dialog_disc_label.ui
-rw-r--r--  root/root   /usr/share/synaptic/gtkbuilder/dialog_download_error.ui
-rw-r--r--  root/root   /usr/share/synaptic/gtkbuilder/dialog_new_repositroy.ui
-rw-r--r--  root/root   /usr/share/synaptic/gtkbuilder/dialog_update_outdated.ui
-rwxr-xr-x  root/root   /usr/sbin/synaptic

Control files: lines which differ (wdiff format)
------------------------------------------------
Depends: libapt-pkg7.0 (>= 1.9.0), libc6 (>= 2.38), libgcc-s1 (>= 3.0), libgdk-pixbuf-2.0-0 (>= 2.22.0), libglib2.0-0t64 (>= [-2.12.0),-] {+2.40.0),+} libgtk-3-0t64 (>= 3.21.5), libpango-1.0-0 (>= 1.14.0), libstdc++6 (>= [-13.1),-] {+14),+} libvte-2.91-0 (>= 0.49.92), {+libxapian30 (>= 1.4.28~),+} hicolor-icon-theme, [-polkitd | policykit-1,-] {+polkitd,+} pkexec | [-policykit-1-] {+systemd+}
Installed-Size: [-7774-] {+7776+}
Version: [-0.91.6-] {+0.91.7+}
```

/cc @andy128k

PR #237, opened 2026-07-13
workflow: add clean deb build

mvo5:
Build on top of #236 

---

We want a test that ensure correct build-deps in CI, this PR adds it.


andy128k:
Using `apt-get build-dep` in CI is like a chicken and egg problem. It pins dependencies and makes adding new ones (e.g. sqlite in one of PRs) a bit strange.

PR #239, opened 2026-07-13
Add missing const to getName(), getSubViews() and getViews()

amaa-99:
Added the missing const attributes and revised the implementation of the respective methods.
(closes amaa-99/synaptic#128)

### DIFF
--- a/common/rpackagelister.cc
+++ b/common/rpackagelister.cc
@@ -161,15 +161,17 @@ void RPackageLister::setView(unsigned int index)
       _selectedView = _views[0];
 }
 
-vector<string> RPackageLister::getViews()
+vector<string> RPackageLister::getViews() const
 {
    vector<string> views;
-   for (unsigned int i = 0; i != _views.size(); i++)
-      views.push_back(_views[i]->getName());
+   views.reserve(_views.size());
+   for (const RPackageView *view : _views) {
+      views.push_back(view->getName());
+   }
    return views;
 }
 
-vector<string> RPackageLister::getSubViews()
+vector<string> RPackageLister::getSubViews() const
 {
    return _selectedView->getSubViews();
 }
--- a/common/rpackagelister.h
+++ b/common/rpackagelister.h
@@ -212,8 +212,8 @@ class RPackageLister {
    }
 
    void setView(unsigned int index);
-   std::vector<std::string> getViews();
-   std::vector<std::string> getSubViews();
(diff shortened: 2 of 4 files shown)


amaa-99:
Those methods could eventually be inlined, but I'd leave that for another pull request.

mvo5:
> Those methods could eventually be inlined, but I'd leave that for another pull request.

No strong opinion here but is manual inling still worth it? My understanding is that modern compilers are quite good at figuring this out on their own?

common/rpackagelister.cc
```diff
@@ -161,15 +161,16 @@ void RPackageLister::setView(unsigned int index)
       _selectedView = _views[0];
 }
 
-vector<string> RPackageLister::getViews()
+vector<string> RPackageLister::getViews() const
 {
-   vector<string> views;
-   for (unsigned int i = 0; i != _views.size(); i++)
-      views.push_back(_views[i]->getName());
+   vector<string> views(_views.size());
```
mvo5: Hm, this does not look quite correct - AIUI this will create a vector of the given size with empty string and then in the loop we push the actual strings *after* the empry one, no? so this needs either be: views[i] = view->getName() or we create the vector with a capacity of _views.size() (so views.reserve(_views.size());)

Or (my preferred approach actually - we just add the consts and leave these functions as close as possobile as they are (using const_iter maybe?). Something like:
```c++
  -   for (map<string, vector<RPackage *> >::iterator I = _view.begin();
  +   for (map<string, vector<RPackage *> >::const_iterator I = _view.begin();
```
or 
```c++
  -   for (map<string, vector<RPackage *> >::iterator I = _view.begin();
  +   for (auto I = _view.begin();
...
```

amaa-99: Thanks for noticing, that should definitely be a reserve(). Going fix that.

amaa-99:
> > Those methods could eventually be inlined, but I'd leave that for another pull request.
> 
> No strong opinion here but is manual inlining still worth it? My understanding is that modern compilers are quite good at figuring this out on their own?

For inlining in those cases the function definition needs to be in the header file.
The inlining keyword though just increases the probability of actual inlining, but as much as I know the compilers still take it into consideration.
There are some optimizations that the compilers can do at linking time, but it won't go as far.

mvo5 (approved):
Love it! Thank you!

PR #240, opened 2026-07-13
meson: copy gtkbuilder ui files into the build dir

mvo5:
When synaptic was running from the build-dir with autotools it would automatically pick the gtkbuilder files using a build-in fallback (looking into ./gtkbuilder). When we moved to meson we lost this.

So this commit adds it back by ensuring that the *.ui files get installed into the build dir (and from there into the datadir).

With that `ninja -C build && ./build/gtk/synaptic` is possible.

/cc @andy128k

### DIFF
--- a/gtk/gtkbuilder/meson.build
+++ b/gtk/gtkbuilder/meson.build
@@ -1,0 +1,43 @@
+# Copy the .ui files into the build dir so that running synaptic from
+# build/gtk/ picks them up via the "gtkbuilder/" development fallback
+# in rggtkbuilderwindow.cc (and install them from here as well).
+fs = import('fs')
+
+gtkbuilder_ui_files = [
+  'dialog_authentication.ui',
+  'dialog_changelog.ui',
+  'dialog_change_version.ui',
+  'dialog_conffile.ui',
+  'dialog_quit.ui',
+  'dialog_task_descr.ui',
+  'dialog_unmet.ui',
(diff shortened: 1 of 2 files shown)


andy128k:
Honestly I don't know why to bother. There could be other files which require installing an application before running (e.g. icons). Build directory and install directory even have different structures.

I'd rather put those files into [a resource](https://docs.gtk.org/gio/struct.Resource.html) and embed it to a binary. It's even kinda a recommended way to deal with these files.

mvo5:
@andy128k Thanks! I like the suggestion of putting them into a resource! I bother because I'm a bit set in my way and ran `(make && (cd gtk ; ./synaptic) ` for the last couple of years :) So unless anything is wrong I will merge (maybe as an intermediate step) I will probably merge. Wdyt? But resource is very cool.

mvo5:
I created https://github.com/mvo5/synaptic/pull/242 as an alternative - actually much nicer.

andy128k:
@mvo5 Oopsie https://github.com/mvo5/synaptic/pull/243

mvo5:
> @mvo5 Oopsie #243

Haha - well - I will take yours, it was your idea afterall :)

amaa-99:
@mvo5 Having those on the build dir would simplify the debugging (in particular in the IDEs).
I'm already doing that on my work copies, and was planning to make a pull for that at some point.

PR #241, opened 2026-07-13
meson: add new `lint` target

mvo5:
This commit adds a new `lint` target that initially just checks for trailing whiespaces. It is also wired up to the GH CI.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,27 @@
+name: Lint
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


andy128k:
How about getting rid of all of them at once? https://github.com/mvo5/synaptic/pull/187

mvo5:
> How about getting rid of all of them at once? #187

Good point :) Sorry, I replied in there. I still thing we need something like this to ensure we do not regress, even with #187 or somehting similar merged.

PR #242, opened 2026-07-13
gtk: tweak gtkbuilder UI files as resources

mvo5:
A small follow-up #243 that adds the gtkbuilder ui files
as resources:
- remove now-unused SYNAPTIC_GTKBUILDERDIR define
- drop stale "// for development" comment (referred to the removed
  local-file fallback)
- remove unused builder_status variable
- say "resource" instead of "file" in the loading warnings
- add missing trailing newline in gtk/meson.build

/cc @andy128k


andy128k (approved):

amaa-99:
@mvo5 Sorry for my late post here.
Currently the application crashes (i.e. closes not so gracefully) in case it can't find/load the .ui files...
Do you think that could be handled more gracefully (perhaps a standard popup info and graceful shutdown)?
(Very minor, mostly can only occur in a dev. env. I guess. ;))

mvo5:
@amaa-99 Thanks for your feedback - is this still an issue with this merge? Before it would load them dynamically and crash. But now they are compiled in so this should not happen anymore, no?

PR #243, opened 2026-07-13
gtk: include gtkbuilder UI files as resources

andy128k:
This PR implements the gtk resources for the gtkbuilder ui files. This avoids having to install them and allow easy local development (as the ui files become just part of the build "synaptic" binary now and run from everywhere).


mvo5 (approved):
Love it!

PR #244, opened 2026-07-14
gtk: fix lock-version and mark-auto-installed

### LINKED ISSUE #184: Can't mark package automatically installed
I've built Synaptic from a16dbb2dd72163055ddce0f50b37d3d61144d4f2 and noticed that marking packages as automatically installed does nothing. Looks like this feature broke at some point, it was working in Synaptic as shipped in Ubuntu 24.04.

mvo5:
When a stateful GAction without a parameter type is activated from a menu, the activate handler receives parameter == NULL, so reading the new value via g_variant_get_boolean(parameter) always yielded FALSE. As a result both toggles could only ever unset their flag.

Read the actions current state instead and toggle it, and update the state so the menu checkmark reflects the change immediately.

Fixes: #184

### DIFF
--- a/gtk/rgmainwindow.cc
+++ b/gtk/rgmainwindow.cc
@@ -514,8 +514,13 @@ void RGMainWindow::cbMenuAutoInstalledClicked(GSimpleAction *action,
    RGMainWindow *me = (RGMainWindow *) data;
    if (me->_blockActions)
       return;
-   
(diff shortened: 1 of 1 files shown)


nazar-pc (approved):
I gave it a try and can confirm it fixes https://github.com/mvo5/synaptic/issues/184

PR #245, opened 2026-07-14
docs: update INSTALL for meson

mvo5:
We no longer use autotools to build synaptic. So update the INSTALL file to reflect this.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -1,368 +1,65 @@
 Installation Instructions
 *************************
 
-   Copyright (C) 1994-1996, 1999-2002, 2004-2016 Free Software
-Foundation, Inc.
-
-   Copying and distribution of this file, with or without modification,
-are permitted in any medium without royalty provided the copyright
-notice and this notice are preserved.  This file is offered as-is,
-without warranty of any kind.
-
-Basic Installation
-==================
-
-   Briefly, the shell command './configure && make && make install'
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -1,368 +1,66 @@
...
-notice and this notice are preserved.  This file is offered as-is,
-without warranty of any kind.
+Requirements
+============
 
-Basic Installation
-==================
+ * meson and ninja
+ * a C++17 compiler (g++ or clang++)
+ * GTK 3 (>= 3.4), GLib, Pango
```
andy128k: 1. nitpick: It's Gtk+ for versions below 4.
2. Both GLib and Pango are dependencies of Gtk+ so maybe not worth to be mentioned

INSTALL
```diff
@@ -1,368 +1,66 @@
...
+ * libapt-pkg development headers (libapt-pkg-dev)
+ * polkit (libpolkit-gobject-1-dev)
+ * libx11-dev
```
andy128k: AFAIK this one was needed for embedding widgets which does not work since the migration to Gtk+3.

amaa-99:
With the autotools this file was getting (automatically) 'updated' each time the project got configured/built.
It would be great if it wouldn't do that (might be that that's already the case)...

PR #247, opened 2026-07-14
Remove the application path definitions from the generated gcc commands:

amaa-99:
Removed the generation of the respective gcc command options from the master meson.build file.
(closes amaa-99/synaptic#130)


amaa-99:
These were related to the autotools philosophy of changing the installation prefix at build time and aren't working correctly with meson/ninja.

andy128k:
~~Some of them will be needed later. Currently paths are combined in two  places (meson and code), while best practice is to define them once in meson and pass through flags/config.h.~~

amaa-99:
Replaced by https://github.com/mvo5/synaptic/pull/249.

PR #249, opened 2026-07-14
Use prefix option consistently

andy128k:
- Use prefix option consistently. It should always be joined with other paths. So, it's done once in the main meson script and variables with prefix are used elsewhere.
- Define `pkg_datadir` (`PACKAGE_DATA_DIR` in code) for package's specific data (in opposite to global `datadir`).

Supersedes #247


meson.build
```diff
@@ -153,13 +148,6 @@ configure_file(output: 'config.h', configuration: conf)
 
 base_cpp_args = [
   '-DHAVE_CONFIG_H',
-  '-DGETTEXT_PACKAGE="synaptic"',
```
amaa-99, line 156: Removing these helps in the integration of the project on various IDEs (only the values set in the config.h are used in those cases). 👍

andy128k: Next step should be to remove HAVE_CONFIG_H.

mvo5 (approved):
Thank you! Love it!

PR #250, opened 2026-07-15
Remove the obsolete intl source folder

amaa-99:
This code doesn't seem to be referenced/built anywhere in the project.
(closes amaa-99/synaptic#132)


amaa-99:
@mvo5 Could you confirm whether this is somehow used by something?

amaa-99:
@mvo5 I'm going to be away for a couple of weeks... Do you think we can get through the open pulls until Friday?

andy128k:
+1

These days GLib supplies i18n API. So no need to maintain own copy of it.

mvo5 (approved):

mvo5:
> @mvo5 I'm going to be away for a couple of weeks... Do you think we can get through the open pulls until Friday?

I can try but don't worry, anything open I/we can also tweak and merge while you are away. I can push to your branch and just tweak things as needed (that does not change the commit author so its still attributed to you).

PR #251, opened 2026-07-16
Fix memory leaks in RPatternPackageFilter::addPattern()

amaa-99:
Revised the implementation and added the missing freeing of allocated memory.
(closes amaa-99/synaptic#133)


mvo5 (approved):

mvo5:
Thanks!

PR #252, opened 2026-07-16
Fix memory leak in get_gdk_pixbuf()

amaa-99:
Changed the gtk_icon_theme_load_icon() call to not retrieve an error object, thus avoiding the memory leak.
(closes amaa-99/synaptic#134)

### DIFF
--- a/common/i18n.h
+++ b/common/i18n.h
@@ -20,8 +20,7 @@
  * USA
  */
 
-#ifndef _I18N_H_
-#define _I18N_H_
+#pragma once
 
 #include "config.h"  // IWYU pragma: associated
 
@@ -50,7 +49,4 @@
 # define _(String) (String)
 # undef  N_
 # define N_(String) (String)
-#endif
-
-
 #endif
--- a/common/raptoptions.h
+++ b/common/raptoptions.h
@@ -20,8 +20,7 @@
  * USA
  */
 
-#ifndef _RAPTOPTIONS_H_
-#define _RAPTOPTIONS_H_
+#pragma once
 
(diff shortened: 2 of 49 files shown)


amaa-99:
Note: GtkIconLookupFlags defines the bits of the flags bitfield, so the value 0 (no flags set) there is fine.
(Error in the declaration of that gtk function.)

gtk/rgutils.cc
```diff
@@ -37,21 +37,15 @@
...
+// Use g_object_unref() to release your reference to the icon. Returns NULL in case of error.
+GdkPixbuf *get_gdk_pixbuf(const gchar *name, int size)
 {
-   GtkIconTheme *theme;
-   GdkPixbuf *pixbuf;
-   GError *error = NULL;
-
-   theme = gtk_icon_theme_get_default();
-   pixbuf = gtk_icon_theme_load_icon(theme, name, size, 
-				     (GtkIconLookupFlags)0, &error);
```
mvo5: Hm, but don't we want to keep error? And just free it after? Also this contains some whitespace changes that make reading the diff slightly harder, it would be nice to avoid those for easier review (fwiw, I think eventually we move to the `fmt` c-lang-format that andy is suggesting then the whole whitespace thing becomes a non-issue :)

andy128k: Technically handling all errors is a way to go. But in this particular case we are loading images supplied with the application itself and they suppose to be always present and be loadable.

Another option would be to get rid of this code altogether. See #257

amaa-99:
Dropping these for now. Will re-implement after I'm back if it's still a topic.

PR #253, opened 2026-07-16
Adjust the compiler flags in the meson script

amaa-99:
Set the builttype to plain, to avoid Meson's defaults.
Reset the warning level to the one used with automake.
Set the c++ standard to c++17.
(Re)enabled the optimization options.

(closes amaa-99/synaptic#135)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,10 +3,17 @@ project(
   ['c', 'cpp'],
   version: '0.91.7',
   default_options: [
-    'warning_level=1',
-    'cpp_std=gnu++17',
-  ],
+    # The buildtype=plain option in Meson disables default compiler flags, allowing you to
+    # specify custom build flags for total control over the build process.
+    # This is often used by distro packagers to ensure consistent and controlled builds.
+    'buildtype=plain',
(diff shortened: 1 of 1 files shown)


amaa-99:
@mvo5 Do you think we could bump the c++ standard to c++20 or c++23?
(I guess you have a better overview of where do this have to be compilable :wink:)

amaa-99:
@mvo5 My plan would be to enable more and more of the warnings in a granular way as I go through the code issues.
For now I'd keep it as before the meson migration, so that we can easily see if new warnings pop in.
p.s. Meson doesn't have granular control for warnings or optimization options (I guess it will eventually get there, after they sift through their 1.8K open issues). I therefore took the approach of explicitly adding whatever is wanted/needed and suppressing as much as practical the defaults injected by the python tool.

andy128k:
What's the point of `O3` and `lto`? This is just a GUI application where heavy-lifting happens in libraries.

amaa-99:
Dropping this for now (will keep using optimizations on my fork though - it cuts down on the UI latencies.

PR #254, opened 2026-07-16
data: drop `_` from translatable strings

mvo5:
The intltool underscore convention is not understood by gettext's msgfmt --xml, which meson's i18n.merge_file uses. The merged files shipped literal <_message>/<_name> tags: polkit ignored the pkexec authentication message, the metainfo failed appstream validation, and no translations were merged into either file.


mvo5:
This is now obsolete with https://github.com/mvo5/synaptic/pull/269/changes

PR #200, opened 2026-06-30
Port to Gtk4

andy128k:
A port to Gtk4. It already compiles and runs. There are still a lot of things not working.

The branch includes:
- A switch from Autotools to Meson (is needed to add a subproject with [an implementation of std::execution](https://github.com/bemanproject/task) from C++26.
- Usage of coroutines to keep code as close as possible to original one,

PR #257, opened 2026-07-16
Refer to icons by names

andy128k:
Do not load icons into pixbufs and allow Gtk to manage them.

The code is back-ported from Gtk4 branch #200

Supersedes #252

### DIFF
--- a/gtk/gtkpkglist.cc
+++ b/gtk/gtkpkglist.cc
@@ -30,11 +30,6 @@
 #include <apt-pkg/strutl.h>
 #include <cassert>
 #include <cstddef>
-#include <gdk-pixbuf/gdk-pixbuf.h>
-#include <gdk/gdk.h>
-#include <glib-object.h>
-#include <glib.h>
-#include <glib/gtypes.h>
 #include <gtk/gtk.h>
 #include <iostream>
 #include <string>
@@ -240,8 +235,8 @@ static void gtk_pkg_list_init(GtkPkgList *pkg_list)
 {
    //cout << "list_init()" << endl;
    pkg_list->n_columns = N_COLUMNS;
-   pkg_list->column_headers[0] = GDK_TYPE_PIXBUF;
-   pkg_list->column_headers[1] = GDK_TYPE_PIXBUF;
+   pkg_list->column_headers[0] = G_TYPE_STRING;
+   pkg_list->column_headers[1] = G_TYPE_STRING;
    pkg_list->column_headers[2] = G_TYPE_STRING;
    pkg_list->column_headers[3] = G_TYPE_STRING;
    pkg_list->column_headers[4] = G_TYPE_STRING;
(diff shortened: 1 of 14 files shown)


gtk/gtkpkglist.cc
```diff
@@ -472,16 +467,14 @@ gtk_pkg_list_get_value(GtkTreeModel *tree_model,
...
+          const char *icon = RGPackageStatus::pkgStatus.getSupportedIconName(pkg);
+          g_value_set_string(value, icon);
           break;
        }
       case PIXMAP_COLUMN:
        {
-          GdkPixbuf *pix;
-          pix = RGPackageStatus::pkgStatus.getPixbuf(pkg);
-          g_value_set_object(value, pix);
+          std::string icon_name = RGPackageStatus::pkgStatus.getIconName(pkg);
```
mvo5: So gtk_pkg_list_get_value() gets called a lot (really a lot at least when I wrote it maybe gtk changed and its less so now idk). So as written this is a new std::string allocation per call, I think we don't need this, we could have something like std::string names[N_STATUS_COUNT] (similar to the previous StatusPixbuf array) and then just return "const char *" into that. This would avoid the allocation make it very cheap again. Normally I would not care about one allocation but this is (or used to be) a very hot path.

gtk/gtktagtree.cc
```diff
@@ -436,9 +436,8 @@ gtk_tag_tree_get_value(GtkTreeModel *tree_model,
        {
           if (pkg == NULL)
              return;
-          GdkPixbuf *pix;
-          pix = RPackageStatus::pkgStatus.getPixbuf(pkg);
-          g_value_set_object(value, pix);
+          std::string icon_name = RPackageStatus::pkgStatus.getPixbuf(pkg);
```
mvo5: This would not compile, its behind an #ifdef maybe/probably dead code by now but might be worth investigating.

gtk/rggtkbuilderwindow.cc
```diff
@@ -225,14 +224,14 @@ bool RGGtkBuilderWindow::setTextView(const char *widget_name,
...
-bool RGGtkBuilderWindow::setPixmap(const char *widget_name, GdkPixbuf *value)
+bool RGGtkBuilderWindow::setPixmap(const char *widget_name, const char *value)
 {
    GtkWidget *pix = GTK_WIDGET (gtk_builder_get_object (_builder, widget_name));
    if (pix == NULL) {
       cout << "textview == NULL with: " << widget_name << endl;
       return false;
    }
-   gtk_image_set_from_pixbuf(GTK_IMAGE(pix), value);
+   gtk_image_set_from_icon_name(GTK_IMAGE(pix), value, GTK_ICON_SIZE_BUTTON);
```
mvo5, line 234: I'm not sure but I think this changes the icon size, the old code was using 22px (initPixbufs) and this is 16px now(?). But maybe we always scaled it down already so it does not matter, idk. Maybe worth a comment in the commit message.

mvo5 (commented):
Thank you! I love the direction, not having to do this manually anymore is great. Some small inline comments.

gtk/rggtkbuilderwindow.cc
andy128k: Icons are 16x16 so, no visible change here.

gtk/gtktagtree.cc
andy128k: fixed

gtk/gtkpkglist.cc
andy128k: I've changed this to pre-allocated strings. BTW, this is a hot path because GtkTreeModel eagerly creates all objects, Gtk4 has GtkColumnView which renders just a visible part of the list so, reduces allocations to a minimum.

gtk/rggtkbuilderwindow.cc
andy128k: If you mean "tango-icons" which are indeed 22x22, then I don't see a way why they should be displayed. I see "hicolor" ones all the time. They seem to not be used anywhere in the code.

gtk/gtkpkglist.cc
mvo5: Ohhhhh, I love to hear that this is better now in gtk4, I had a really hard time with the old way (it was like this from gtk1-3 iirc).

PR #259, opened 2026-07-17
Replace include guards with #pragma once

andy128k:
Replace include guards with `#pragma once` as was suggested in https://github.com/mvo5/synaptic/pull/187


mvo5 (approved):
\o/

PR #261, opened 2026-07-19
Remove unused code related to window embedding

andy128k:


mvo5 (approved):
Slightly sad about this one, I remember when it was used as part of an installer - oh well - not how one would do these things nowdays anymore :)

PR #246, opened 2026-07-14
Invalid non-blank line in synaptic.desktop.in

amaa-99:
Removed the leading '_' from the Name, GenericName and Comment entries.
(closes amaa-99/synaptic#129)

PR #263, opened 2026-07-21
Revert "Invalid non-blank line in synaptic.desktop.in:"

AsciiWolf:
This reverts commit 4e7668fad3c778ba13220e666abc8c9ddb8623e3.

The "_" prefixed lines in *.in files are automatically replaced with the translated ones by gettext. Removing them will mark these strings as non-translatable.

It is **correct** that `msgfmt` complains about them since *.in suffixed files are template files and are **not supposed** to be validated or installed directly. This is automatically handled during the compilation process, but I am not sure if it still works since you replaced autotools with meson recently. If it does not work, you have to **fix it** in the meson.build file - that's the correct way.

/cc @mvo5 @amaa-99

### DIFF
--- a/data/synaptic.desktop.in
+++ b/data/synaptic.desktop.in
@@ -1,7 +1,7 @@
 [Desktop Entry]
-Name=Synaptic Package Manager
-GenericName=Package Manager
-Comment=Install, remove and upgrade software packages
+_Name=Synaptic Package Manager
+_GenericName=Package Manager
+_Comment=Install, remove and upgrade software packages
 Exec=synaptic-pkexec
 Icon=synaptic
 Terminal=false


mvo5:
Hey @AsciiWolf ! Nice to see you and thanks for thanks for the PR! I was under the impression that the meson integration of gettext would no longer need the "_" - did I misread/misassume this? If so I'm of course happy to merge this.

@AsciiWolf if you have a bit of time I would love if you could play around a bit with the latest git version of synaptic - if all is well I would love to make a release soon(ish).

AsciiWolf:
>I was under the impression that the meson integration of gettext would no longer need the "_"

Oh! It is possible, I am not an expert on meson. :) This needs to be tested (I can do it in upcoming weeks, but unfortunately not sooner; I will also test the git version of Synaptic as part of this).

AsciiWolf:
By the way, what about the AppStream metainfo file? It [also has](https://github.com/mvo5/synaptic/blob/master/data/io.github.mvo5.synaptic.metainfo.xml.in#L7) the "_" characters.

AtesComp:
FYI:
meson i18n gettext integration how to mark strings for translation in .in files

In Meson and GNU gettext, how you mark strings in .in files depends on the file type. For Desktop/Metainfo files, prefix the keys with an underscore _. For XML/GSettings files, add translatable="yes". Meson’s [i18n module](https://mesonbuild.com/i18n-module.html) automatically handles extraction.

1. Desktop and Metainfo Files (.desktop.in & .metainfo.xml.in)
Prefix the translatable property or key with a single underscore _.
Example (app.desktop.in):
```ini
[Desktop Entry]
Type=Application
_Name=My Application
_GenericName=Text Editor
_Comment=Edit your text files
Exec=app
```
Use code with caution.

2. XML Files (.xml.in)
For standard XML files, strings are marked using translatable="yes" attributes, often alongside an instruction set (ITS) file for gettext to parse them properly.
Example (data.xml.in):
```xml
<?xml version="[1](https://github.com/mesonbuild/meson/issues/1565).0" encoding="UTF-8"?>
<component>
  <id>com.example.App</id>
  <name translatable="yes">My Application</name>
  <summary translatable="yes">A simple app summary</summary>
</component>
```
Use code with caution.

3. GSettings Schemas (.gschema.xml)
For schema files, use the translatable="yes" attribute on the \<default>, \<summary>, or \<description> tags.
Example (org.example.app.gschema.xml):
```xml
<?xml version="[1](https://discourse.gnome.org/t/xgettext-dont-extract-strings-from-ui-files/25063).0" encoding="UTF-8"?>
<schemalist>
  <schema id="org.example.app" path="/org/example/app/">
    <key name="greeting" type="s">
      <default translatable="yes">Hello World</default>
      <summary translatable="yes">A greeting message</summary>
      <description translatable="yes">The message shown to users on startup.</description>
    </key>
  </schema>
</schemalist>
```
Use code with caution.

4. Meson Build Configuration (meson.build)
To process these files and extract the marked strings, use i18n.merge_file() inside your Meson Build System configuration.
Example (meson.build):
```meson
i18n = import('i18n')

# Desktop file
i18n.merge_file(
  input: 'data/app.desktop.in',
  output: 'app.desktop',
  type: 'desktop',
  po_dir: meson.project_source_root() / 'po',
  install: true,
  install_dir: get_option('datadir') / 'applications'
)

# Metainfo file
i18n.merge_file(
  input: 'data/app.metainfo.xml.in',
  output: 'app.metainfo.xml',
  po_dir: meson.project_source_root() / 'po',
  install: true,
  install_dir: get_option('datadir') / 'metainfo'
)
```

mvo5:
Thanks for the detailed examples. So I looked at the desktop file and it seems its getting it right and this revert is not needed? Maybe I'm missing something but I did (in 
```
$ meson build
$ ninja -C build synaptic-pot
$ grep   -A3 "synaptic\.desktop\.in" po/synaptic.pot 
#: gtk/rgmainwindow.cc:2312 data/synaptic.desktop.in:3
#: data/io.github.mvo5.synaptic.metainfo.xml.in:7
msgid "Synaptic Package Manager"
msgstr ""
--
#: data/synaptic.desktop.in:4
msgid "Package Manager"
msgstr ""

#: data/synaptic.desktop.in:5 data/io.github.mvo5.synaptic.metainfo.xml.in:11
msgid "Install, remove and upgrade software packages"
msgstr ""
```

After applying the fixes from #269 - I think we are good and we don't need this PR. 
The way meson extracts is plain xgettext/msgmerge, the previous autotools used
intltool, they just have different conventions: intltool needs the "_" but xgettext is
smater and knows the standard desktop keys so no need to mark e.g. name.

But let me know if I am missing something here, i18n is always a bit tricky :)

AsciiWolf:
Looks good! Just make sure it is a new template, not a refreshed old one (that can potentially contain old strings).

amaa-99:
Hi everyone. Does anyone know how's it looking regarding the failed tests mentioned on my original pull?
https://github.com/mvo5/synaptic/pull/246
@mvo5 Could those tests be added/enabled, or are those completely inadequate (and shouldn't be used)?
`msgfmt --desktop --template=../data/synaptic.desktop.in -o synaptic.desktop -d ../po -v --check`
(Sorry, I just got back and couldn't yet check deeply into the details on this.)

AsciiWolf:
Well, you should ideally verify the actual *.desktop file, not the *.in template.

amaa-99:
If I'm interpreting the `msgfmt` documentation correctly the `--check` option intention is to check the template for issues:

>    Input file interpretation:
>        -c, --check
>               perform all the checks implied by --check-format,
>               --check-header, --check-domain
> 
>        --check-format
>               check language dependent format strings
> 
>        --check-header
>               verify presence and contents of the header entry
> 
>        --check-domain
>               check for conflicts between domain directives and the
>               --output-file option

The last time I ran `msgfmt` with the `--check` option there were quite a few 'issues' reported:
```
...
../po/no.po:2080: number of format specifications in 'msgid_plural' and 'msgstr[0]' does not match
../po/no.po:2196: number of format specifications in 'msgid' and 'msgstr' does not match
msgfmt: found 10 fatal errors
../po/pa.po:7: warning: header field 'Last-Translator' still has the initial default value
../po/ps.po:7: warning: header field 'Last-Translator' still has the initial default value
../po/pt.po:7: warning: header field 'Last-Translator' still has the initial default value
../po/pt.po:1280: number of format specifications in 'msgid' and 'msgstr' does not match
msgfmt: found 1 fatal error
../po/pt_PT.po:7: warning: header field 'Last-Translator' still has the initial default value
../po/pt_PT.po:7: warning: header field 'Language' still has the initial default value
...
```
I think it would be beneficial to have these checks enabled (and eventually even added to the 'lint' checks) so that those issues are detected and clearly visible, but there were still too many issues reported the last time I've checked...

AsciiWolf:
>If I'm interpreting the msgfmt documentation correctly the --check option intention is to check the template for issues

Isn't it about the pot template? That's a different thing. :)

I was talking about *.in files that (can) have placeholder strings intended to be automatically replaced with the correct text by autotools/meson/gettext etc. These should not be passed directly to the validation tools. Instead, the final desktop/xml/etc. files should be validated.

PR #270, opened 2026-07-26
Extract `poll` and `finish` methods of install progress

andy128k:
This PR:
- extracts `poll` method so polling could be done externally (and interspersed with other UI updates if needed).
- adds `finish` method and use polymorphism instead of preprocessor.
- removes `goto` by using `std::unique_ptr`.

### DIFF
--- a/common/rinstallprogress.cc
+++ b/common/rinstallprogress.cc
@@ -75,8 +75,6 @@ pkgPackageManager::OrderResult RInstallProgress::start(pkgPackageManager *pm,
                                                        int numPackagesTotal)
 {
    pkgPackageManager::OrderResult res;
-   int ret;
-   pid_t _child_id;
 
    // cout << "RInstallProgress::start()" << endl;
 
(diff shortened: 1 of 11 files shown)


common/rinstallprogress.cc
```diff
@@ -136,19 +134,22 @@ pkgPackageManager::OrderResult RInstallProgress::start(pkgPackageManager *pm,
...
-      updateInterface();
+   return pkgPackageManager::OrderResult::Incomplete;
+}
 
-   res = (pkgPackageManager::OrderResult)WEXITSTATUS(ret);
+pkgPackageManager::OrderResult RInstallProgress::poll()
+{
+   int ret;
+   if (waitpid(_child_id, &ret, WNOHANG) == 0) {
+      return pkgPackageManager::OrderResult::Incomplete;
```
mvo5: So pkgPackageManager::OrderResult::Incomplete traditionally means that a media-swap (cdrom) is required. Those are of course nowdays extremly rare and maybe we should drop that but AIUI this change would break this support. I tink its fine to just make it `bool RInstallProgress::poll()`

mvo5:
Thanks! There is lots to like here, this is some very old (and inelegant) code so a cleanup/refactor is appreciated (but see my inline comment)

common/rinstallprogress.cc
andy128k: Well, does it really matter why the process is "incomplete"? I mean it could be waiting for a next CD/DVD or a next batch downloaded from internet. In both cases reaction is the same -- continue polling.

OTOH If you think that these two states should be separate, then both `start` and `poll` methods should be changed to return something else (e.g. `std::optional<OrderResult>`).

----

This PR alone will not help to achieve a separation of installation process and UI. Unfortunately the interface of `pkgAcquire` is synchronous. I am afraid the best option would be to move the whole process into an own thread and communicate with UI via channels.

mvo5: So the issue is that on a real media-swap event the exit code is the same (incomplete) and the UI would assume it is finished and keep polling. But incomplete from the package manager is an end-state, i.e. the loop needs to exit and the dialog with "please insert next disk" needs to come up. its all a bit academic of course as there are no cdrom based distros anymore :) 

But I love your suggestions about std::optional - updated the code and pushed here (hope that is okay) and once green will mrege, the refactor itself is very  nice.

PR #150, opened 2025-05-05
Add Deb822 sources support (issue #149)

aybanda:
This PR implements support for the Deb822 source format in Synaptic, as requested in [issue #149](https://github.com/mvo5/synaptic/issues/149).

## What’s new

- Introduces `RDeb822Source` and `RSourceManager` classes for parsing, validating, and serializing Deb822 sources.
- Updates the repository window UI to allow adding, editing, and removing Deb822 sources.
- Handles multiple URIs and suites per source, and supports enabling/disabling sources.
- Ensures backward compatibility with existing sources.

## Testing

- Added C++ integration tests (`tests/test_deb822_integration.cc`) to verify parsing, file I/O, and validation.
- Supplemented with a Python script to simulate and test edge cases for Deb822 parsing and logic. (not included)

### Sample Python test results

Created test directory: /tmp/tmpabcd1234
Initialized source manager
Source validation: PASS
Source string representation:
deb http://example.com stable main
Successfully wrote source to file: /tmp/tmpabcd1234/test.sources
Read source from file:
Type: deb
URI: http://example.com
Suite: stable
Components: main
Disabled source test: PASS
Multiple URIs test: PASS
Multiple suites test: PASS

PR #274, opened 2026-09-01
Rebase Deb822 sources support (#150) onto master; fix build and a stanza-separator data loss

tzh476:
Rebases @aybanda's Deb822 support (#150) onto current master, and makes it build and run.

**@aybanda wrote this feature — I have only rebased it and fixed what the build and tests then
found.** Their PR has been unmergeable for a while (`mergeable_state: dirty`) and they have not been
active since 2025-07, so I did the rebase on a fork rather than touching their branch. If they would
rather carry it themselves, please close this and I will hand over the commits.

@mvo5, this addresses your request from 2025-05-13:

> I still get compiler errors when I run this locally, if you could rebase the branch to current
> master that would be great, this way we will get working CI again

**And the empty-repositories-window you reported does not reproduce.** I fed your exact
`debian.sources` from [this comment](https://github.com/mvo5/synaptic/pull/150#issuecomment-2851244897)
to the real `SourcesList::ReadDeb822SourcePart()`:

```
rec 1: type=1027 URI=http://ftp.de.debian.org/debian/ Dist=trixie NumSections=2 [main] [non-free-firmware]
rec 2: type=1025 URI=http://security.debian.org/debian-security/ Dist=trixie-security NumSections=2 [main] [non-free-firmware]
TOTAL RECORDS: 2
```

`1027` = `Deb|DebSrc` (your `Types: deb deb-src`), `1025` = `Deb`. Both stanzas parse with correct
URIs, suites and components.

### Three commits, each droppable independently

**1. Merge master (16 conflicts, 4 files).** master had modernised the touched files (`std::`
qualification, brace reflow, IWYU includes) while the branch added Deb822, so most resolutions are
master's style with the branch's substance. Two hunks take the branch's version because master's code
is *superseded*, not merely restyled: the `sourcesByFile` write path (keeping master's `filenames`
loop would leave two writers in one function), and the `Deb|DebSrc` OR-ing in the type handling,
which is the feature itself.

**2. Build fixes.** The branch predates the autotools→meson migration, so its `Makefile.am` listed
`rsource_deb822.cc` and `common/meson.build` did not — 8 undefined references at link. Plus
`<sstream>`/`<vector>` now included explicitly in `rgrepositorywin.cc` (they were arriving
transitively, so reordering includes broke `stringstream ss(...)`), and one unqualified
`new string[...]` → `new std::string[...]`.

**3. A whitespace-only line separates deb822 stanzas** — a real data-loss fix, and the one commit
here that changes behaviour, so it is easy to drop if you would rather handle it separately.

`ParseDeb822File` tested `line.empty()`, so a separator line containing spaces counted as content:
parsing ran on into the next stanza and `fields[key] = value` overwrote the previous one. The earlier
source vanished, with no error, and the function still returned `true`.

Two byte-identical inputs differing only in the separator (`\n\n` vs `\n   \n`):

| input | before | after |
|---|---|---|
| blank-line separator | 2 records | 2 records |
| whitespace-only separator | **1 record** | 2 records |
| CRLF file (separator is `\r`) | **1 record** | 2 records |
| your `debian.sources` | 2 | 2 |
| empty / comments-only | 0 / 0 | 0 / 0 |
| `Enabled: no` | honoured | honoured |

CRLF was losing sources for the same reason — `std::getline` leaves `"\r"`, previously treated as
content.

### Verification

Debian bookworm container, meson + `ninja -k 0`, run against **master as the baseline** so the
comparison is checkable:

| | master `8e99020` | this branch |
|---|---|---|
| compile errors | 0 | 0 |
| undefined references | 0 | 0 |
| objects | 57 | 58 |
| `gtk/synaptic` links | yes | yes |
| FAILED targets | 1 | 1 — *the same one* |

The single failure is identical on both sides and environmental:
`msgfmt: cannot locate ITS rules for ../data/com.ubuntu.pkexec.synaptic.policy.in`. It fails on
untouched master in this container too, so it is not from these changes.

### What is NOT verified

- **The GTK dialog itself.** The reader is exercised through the real library; nobody has clicked a
  button. This needs your eyes on a real apt tree.
- **The write / round-trip path.**

### Left alone deliberately

Three things I could have changed and did not, because they are @aybanda's calls, not merge fallout:

- **Multi-value `URIs:` / `Suites:`** keeps only the first value, so
  `URIs: http://a/deb http://mirror/deb` silently drops the mirror. Real, but fixing it means deciding
  how one stanza maps onto N `SourceRecord`s — a design change.
- `type_val.find("deb")` also matches inside `"deb-src"`, so `set_deb` is always true when
  `set_debsrc` is. Possibly intended for `Types: deb deb-src`.
- `rpackagemanager.*`, `rsourcemanager.*` and `rdeb822source.*` (~800 lines) have no callers, so they
  are not in the meson build. Left in the tree rather than deleted.

Also unchanged: an unused local `orig_p` in `rsources.cc`, which is why the warning count goes 27→28.

---

Disclosure: this rebase, the fixes and the test harness were done with AI assistance (Claude). Every
number above comes from a run I can reproduce on request — the harness and the nine `.sources`
fixtures are available if you want them in `tests/`.


tzh476:
Pushed a regression test for the separator fix (`b297d32`), since a fix without a test that can fail
is not worth much.

`tests/test_deb822_separator.cc`, in the same plain-`main()` style as the other tests here.
**Verified both ways:**

```
without the fix:  FAIL whitespace-separator (got 1, expected 2)
                  FAIL tab-separator        (got 1, expected 2)
                  FAIL crlf-separator       (got 1, expected 2)
                  3 of 9 checks failed
with the fix:     all 9 checks passed
```

`blank-separator` is a deliberate control — if it ever fails, the harness is broken and the other
cases prove nothing. The other five checks (single stanza, trailing blank lines, comments-only, empty
file, stanza missing a required field) pin the surrounding behaviour so a later change here cannot
quietly start inventing or dropping records.

Full suite in a Debian bookworm container, from a tree extracted out of the commit itself:

```
1/4 test_deb822_separator OK   0.01s
2/4 test_rpackagefilter   OK   0.02s
3/4 test_rpackageview     OK   0.02s
4/4 test_rpackage         OK   0.04s
Ok: 4   Fail: 0   Timeout: 0
```

One note on the branch's existing `tests/test_deb822_integration.cc`: it is 324 lines and **never
built or run** — it is not referenced from `tests/meson.build`, and it needs gtest, which this project
does not use anywhere else. I left it in place rather than deleting @aybanda's file, but it is not
providing coverage today. Happy to port it to this style in a follow-up if that is useful.

CI here is showing `action_required` on both workflows rather than a result, which I believe is the
first-time-contributor approval gate — `check-runs` is `total_count: 0`, so it has not run rather than
failed. The container numbers above are what I have until someone approves it.

tzh476:
One more finding, on the **write** path this time — reporting rather than changing, since the fix is a
design call.

**Saving a `.sources` file destroys `Architectures`, `Languages`, `Targets` and any comments.**

Reading a realistic file and then calling the same `UpdateSources()` the repository dialog calls
(`gtk/rgrepositorywin.cc:730`), diffed against what the user started with:

```diff
-# Debian main repo, kept as a comment
+Enabled: yes
-Architectures: amd64 arm64
-Languages: en de
-Targets: deb-src
```

Reproduced two independent ways: `ParseDeb822File` → `WriteDeb822File` directly, and the full
`ReadSources()` → `UpdateSources()` path against an apt sandbox.

### Cause: the reader and writer disagree about the field set

| | fields |
|---|---|
| `ParseDeb822File` **reads** | `Types` `URIs` `Suites` `Components` `Signed-By` `Enabled` `Disabled` |
| `WriteDeb822File` **writes** | `Types` `URIs` `Suites` `Components` `Signed-By` **`Architectures` `Languages` `Targets`** |

`Deb822Entry` has `Architectures`, `Languages`, `Targets` and `Comment` members, and the writer
faithfully emits three of them — but the parser never populates any, so they are always `""` by the
time the writer runs. Since `rsources.cc:352` opens the file with `ios::out` (truncating) before the
deb822 branch writes the stanzas back, those fields are not merely un-preserved — they are **erased
from the user's file on the first save**. On a multi-arch box (`Architectures`) or a bandwidth-limited
one (`Languages`) that silently changes what apt does on the next update.

### Two things I checked and they are NOT bugs

I want to be clear about these, because both looked alarming first:

1. **"Saving duplicates every source"** (2 in, 4 out) — my sandbox had
   `Dir::Etc::sourcelist.d` and `Dir::Etc::sourceparts` pointing at the same directory, so
   `ReadSources()` scanned it twice. With a realistic layout it is 2 in / 2 out. My harness's fault.
2. **`ReadSources()` returning false and reading 5 records from a 2-source file** —
   `FindDir("Dir::Etc::sourcelist.d")` resolved to `/`, so the fallback at `rsources.cc:232-234` read
   the *container's own* `/etc/apt/sources.list.d/`. Also my harness.

### Not fixed in this PR

Teaching the parser to read those fields (and to attach comments) is completing the feature rather
than fixing the rebase: it needs new `Deb822Entry` population and decisions about how comments attach
to stanzas — the `Comment` member is currently unused on both sides. That is @aybanda's design space,
so I have left it alone, same as the multi-value `URIs:`/`Suites:` case and `type_val.find("deb")`.

Happy to implement whichever of these you'd like, in whatever shape you prefer — I just did not want
to make those calls inside someone else's feature branch. Harnesses for all of the above are available
if you want them as tests.

tzh476:
Fixed the write-path data loss I reported above, in `cae3afa0` — it turned out to be a one-sided gap
rather than the design question I took it for, so it belongs in this PR after all.

**Saving a `.sources` file was erasing `Architectures`, `Languages` and `Targets` from the user's file.**
`rsources.cc:352` truncates the file before rewriting the stanzas, so those fields were not merely
un-preserved; they were removed on the first save. On a multi-arch machine that silently changes what
apt does on the next `update`.

Everything needed to round-trip them was already here — `Deb822Entry` declares all three,
`WriteDeb822File` emits all three, and `ConvertToSourceRecord`/`ConvertFromSourceRecord` already carry
them through `record.Comment`. Only `ParseDeb822File` never populated them, so they were always empty
strings by the time the writer ran. The fix is those three assignments, at both stanza-emit sites.
No other file changed.

**The test is a real control — same test, only `common/rsource_deb822.cc` swapped:**

```
without the fix:  FAIL architectures / languages / targets   3 of 4 checks failed
with the fix:     all 4 checks passed
```

`tests/test_deb822_roundtrip.cc` asserts on the **saved file text**, not on the parsed structs. That
matters: my first harness compared entry structs and reported "0 fields lost", which was worthless —
it read `Architectures` through the parser on both sides, so both were `''` and compared equal. A
metric that reads its input through the broken stage cannot detect that stage being broken. The
`Components` case is a deliberate positive control: the parser already read that field, so if it ever
fails the harness itself is broken.

**Full suite, Debian bookworm container, `meson test`:** 5/5 OK, 0 build errors.

One thing worth flagging rather than leaving for you to hit: `test_rpackage` SIGSEGVs on the *first*
run against a freshly configured build directory, and passes on every run after that. **This is not
from my change** — I reproduced it on both arms:

```
this branch     fresh build dir -> test_rpackage SIGSEGV;  5 subsequent runs 5/5 OK
b297d32a        fresh build dir -> test_rpackage SIGSEGV;  8 subsequent runs 8/8 OK
```

So it is pre-existing on this branch's head and orthogonal to deb822. I have not investigated further
since it is outside this PR's scope, but it will look like a new failure to anyone who configures a
clean tree, so it seemed worth naming.

Status of the three paths now: reader tested (2 data-loss bugs fixed), writer tested (this one fixed),
GTK dialog still unexercised — that one needs a display and a human clicking the button.

tzh476:
Two more merge blockers cleared in `0f287fa1`. I saw you merged #273 this morning, so flagging these
now rather than letting them sit.

**1. Four `DEBUG:` prints were shipping in the repository dialog.**

```
gtk/rgrepositorywin.cc:425  g_print("DEBUG: Number of source records read into _lst: %lu\n", ...)
gtk/rgrepositorywin.cc:440  g_print("DEBUG: Adding source to display - URI: %s, Type: %s\n", ...)
gtk/rgrepositorywin.cc:464  g_print("DEBUG: Preparing to append to list store for URI: %s\n", ...)
gtk/rgrepositorywin.cc:488  g_print("DEBUG: Successfully set data for URI: %s\n", ...)
```

They are not on master — this branch introduced them — and they print to stdout once per source row on
every open of the dialog. Removed, pure deletions, no logic touched. Checked against the built artifact
rather than the source, since that is what users actually get:

```
strings build/gtk/synaptic | grep -c "DEBUG: "     before: 4     after: 0
```

**2. The dialog population path now has a test — this is the one that answers your original report.**

Your blocking complaint on this feature was functional: *"I only get an empty window when I open the
repository dialog"*. That path was never exercised, so a green build did not answer you. I fed the
`.sources` file from your comment through the same steps `Run()` takes to fill the list —
`ReadDeb822SourcePart`, the `Comment` skip, and the deb/deb-src type-display logic:

```
  mvo5-reported-file:
     row: deb, deb-src  http://ftp.de.debian.org/debian/             trixie
     row: deb           http://security.debian.org/debian-security/  trixie-security
  ok mvo5-reported-file: 2 row(s)
  ok empty-file-is-empty: 0 row(s)
```

**Your file does not produce an empty list on this branch.** The empty-file case is a deliberate
negative control — it yields 0 rows, so the 2 above cannot be an artifact of a counter that always
counts.

Being precise about what this does not cover: the test stops short of constructing the `GtkWindow`,
which needs the builder resources and a real display. It covers the data that decides whether the
window comes up populated, which is what your report is about, but **nobody has actually clicked
through the dialog** — that still needs a human at a display, and I would not claim otherwise.

**Full suite, Debian bookworm container: 6/6 OK, 0 build errors.** Branch is MERGEABLE against current
master; I checked your #273 for overlap and there is none (zero shared files, clean trial merge), so no
rebase is needed from my side.

Where the three paths stand now: reader tested (2 data-loss bugs fixed), writer tested (1 fixed),
dialog population tested, GTK interaction still human-only.

tzh476:
Last piece of the write-path data loss, in `517b93a8`: **comments were also being deleted on save.**

I had reported this one as a design question alongside the fields, and I was wrong about that too —
it is the same one-sided gap. `Deb822Entry` declares `Comment`, and `WriteDeb822File` already emits it
in the right place, immediately before the stanza it belongs to (`rsource_deb822.cc:125`). Only the
parser never populated it: line 66 saw `#` and `continue`d, so the comment was gone before the writer
ran. With `rsources.cc:352` truncating the file first, saving deleted the user's own annotations.

deb822 has no comment syntax of its own, so apt reads `#` lines as belonging to the stanza that
follows. This accumulates them and attaches them to the next stanza completed — matching what the
writer already assumed. A stanza skipped for missing `Types`/`URIs`/`Suites` clears the accumulator,
so an orphaned comment cannot migrate onto an unrelated source.

Three checks added to `tests/test_deb822_roundtrip.cc`, verified both ways:

```
without the fix:  FAIL comment-first-stanza
                  FAIL comment-with-colon
                  FAIL comment-second-stanza
                  3 of 7 failed
with the fix:     all 7 passed
```

The colon case is deliberate — `# See: https://wiki.debian.org/SourcesList` must not be mistaken for
a field. The second-stanza case pins the *attachment rule*, not merely that "a comment survived".

Full suite in a Debian bookworm container, built from a tree extracted out of the commit itself:
**6/6 tests OK, 0 compile errors, 0 undefined references.**

---

So the write path is now complete: `Architectures`, `Languages`, `Targets` and comments all survive a
save, each with a regression test that fails without its fix. Taken together with the reader fixes,
what a `.sources` file loses on save has gone from four things to none that I can find.

@mvo5 — I saw you merged #273 this morning, so you are clearly around; no rush from my side, and I am
happy to squash or reorder any of these if the history is inconvenient. @aybanda's implementation is
still the substance of the PR.

tzh476:
@mvo5 I think I found the cause of the problem you reported on #150, and it is not in the parser.

You wrote on 2025-06-02:

> I did a quick test on an ubuntu test machine and I don't see the content of
> `/etc/apt/sources.list.d/ubuntu.sources` in the rgrepositoryeditor window. So it seems there is still
> a bit of work, ideally we would enumerate everything `*.sources` in the apt `Dir::Etc::sourceparts`

`ReadSources()` was scanning **two** directories:

```cpp
string Deb822Parts = _config->FindDir("Dir::Etc::sourcelist.d");
if (Deb822Parts.empty() || Deb822Parts == "/")
   Deb822Parts = "/etc/apt/sources.list.d/";
...
string Parts = _config->FindDir("Dir::Etc::sourceparts");
```

**`Dir::Etc::sourcelist.d` is not an apt configuration key.** I checked what apt actually returns for
it, with `sourceparts` pointed at a sandbox:

```
FindDir("Dir::Etc::sourcelist.d") = [/]
FindDir("Dir::Etc::sourceparts")  = [/tmp/zz/sources.list.d/]
```

So the first lookup always fell through to the hardcoded `/etc/apt/sources.list.d/` — which is the
directory `sourceparts` already points at on a normal system. Every `.sources` file was read **twice**.
In a sandbox with three `.sources` files and one legacy `.list`:

```
before:  6 records   (each .sources counted twice)
after:   4 records   (3 .sources + 1 .list, each exactly once)
```

Dropping the bogus key leaves the single `sourceparts` scan, which is what you asked for and which
already handles both deb822 and one-line files.

While there I also fixed a buffer under-read next to it — the extension test was
`strcmp(d_name + strlen(d_name) - 8, ".sources")`, which reads in front of the buffer for any filename
shorter than `.sources`. A two-character file in `sources.list.d` is enough to trigger it. The new test
puts `ab` and `x` in the directory so that path is actually exercised.

`tests/test_deb822_sourceparts.cc` covers the enumeration itself. It is a real control — same test,
only `common/rsources.cc` swapped:

```
without the fix:  FAIL, got 6 records
with the fix:     ok, 4 records, no duplicates
```

**Full suite in a Debian bookworm container: 7/7 OK, 0 build errors.**

One caveat I want to be straight about: this explains a *duplicate* listing, and your report was an
*empty* window. Those are different symptoms, so I am not claiming this closes your report outright —
the two-directory scan is a real bug either way, and I would rather you know the mechanism than take a
"fixed!" from me. If you still see an empty window after this, the remaining suspect is the apt
sandbox/config in your test environment rather than the directory walk, and I am happy to dig further.

tzh476:
Follow-up with a stronger result, and a narrowing of my own claim.

**The duplicate scan reproduces on a real, unmodified Debian — not just in my sandbox.** I ran the
actual `ReadSources()` inside a stock `debian:bookworm` container, nothing configured, which happens to
be exactly the "modernized" layout you described (`/etc/apt/sources.list` does not exist; only
`/etc/apt/sources.list.d/debian.sources`):

```
UNFIXED                                        FIXED
row: deb.debian.org/debian          bookworm   row: deb.debian.org/debian          bookworm
row: deb.debian.org/debian-security bookworm-… row: deb.debian.org/debian-security bookworm-…
row: deb.debian.org/debian          bookworm            (2 rows)
row: deb.debian.org/debian-security bookworm-…
         (4 rows — every source twice)
```

Same binary, same machine, only `common/rsources.cc` swapped. So on any system where
`Dir::Etc::sourceparts` resolves to `/etc/apt/sources.list.d/` — i.e. the default — the old code listed
every repository twice.

**And your exact file parses correctly.** I took the `debian.sources` from your 2025-05-05 comment
verbatim, including the trailing spaces after `Components`, and drove it through `ReadSources()` (not
the single-file helper — the same call the dialog makes):

```
row: type=1027 URI=http://ftp.de.debian.org/debian/          Dist=trixie          sections=2
row: type=1027 URI=http://security.debian.org/debian-security/ Dist=trixie-security sections=2
ReadSources() returned true, 2 row(s) would be shown
```

`type=1027` is `Deb|DebSrc|Deb822`, matching your `Types: deb deb-src`.

**Where that leaves your report — narrowing rather than claiming.** I cannot reproduce an *empty*
window from the sources code: your file yields rows, and a default Debian yields rows. What I found and
fixed is duplication, which is a different symptom. Two things I can rule out and one I cannot:

- Not the parser: your file produces 2 correct records.
- Not the directory walk: a stock system enumerates `sources.list.d` fine.
- **Cannot rule out** the GTK layer itself — everything above stops at the record list, and I have no
  display to drive `RGRepositoryEditor::Run()` end to end. If the rows exist but the tree view shows
  nothing, the fault is between `gtk_list_store_append` and the widget, which needs someone at a screen.

Given your 2025-05-05 note that it also *"fails to build in the new handleFailedInstallation"*, it is
also possible the empty window you saw came from a partially-built tree rather than from this code. The
branch builds clean now (7/7 tests, 0 errors in a bookworm container), so it would be worth one more
look on your side before assuming the symptom survives.

tzh476:
Dropped six files from this PR in `b517c578` — one of which you had already deleted yourself, @mvo5.

The rebase was silently reintroducing `common/rpackagemanager.h`, which you removed in `c46ca917`
*"Remove the orphaned rswig.cc, rswig.h, rpackagemanager.h and rdependency.h files"* (2026-07-04).
A rebase should not resurrect a file the project deliberately deleted, so that is on me for not
checking the merge result against your recent history.

All six were unbuilt and unreferenced:

| files | class | references outside its own pair |
|---|---|---|
| `common/rpackagemanager.{cc,h}` | `RPackageManager` | **0** |
| `common/rsourcemanager.{cc,h}` | `RSourceManager` | **0** |
| `common/rdeb822source.{cc,h}` | `RDeb822Source` | **0 includes anywhere** |

None appear in `common/meson.build`, so none were ever compiled.

`rdeb822source.h` was worse than dead weight: it declares a **second, different class also named
`RDeb822Source`**, distinct from the one in `rsource_deb822.h` that the code actually uses. Every live
include is `rsource_deb822.h` — `rsources.cc:31`, `gtk/rgrepositorywin.cc:34`, `rsource_deb822.cc:6`,
`tests/test_deb822_roundtrip.cc:3`. Two conflicting declarations of one class name is a hazard waiting
for someone to add either file to the build.

**−1013 lines.** The PR is now 16 files / +1702 instead of 22 / +2715.

Verified against a baseline in a Debian bookworm container, both directions:

```
with the six files present:  8/8 tests OK, 0 errors, 0 undefined refs
with them removed:           8/8 tests OK, 0 errors, 0 undefined refs
```

One thing I want to flag rather than hide: on one run of the reduced tree, `test_rpackage` aborted
with **SIGSEGV**. It is not from this change — `test_rpackage` references none of the removed files and
links only `libsynaptic` — and it passed 3/3 on re-run plus two clean full-suite passes on the same
tree. It reads the container's live apt cache, which is the likely source of the flake. I mention it
because "it passed the second time" is exactly the kind of thing worth saying out loud, not quietly.

Disclosure: this work was done with AI assistance (Claude). Every number above is from a run I can
reproduce on request.

tzh476:
Also dropped `tests/test_deb822_integration.cc` in `27eb6255`. It is 324 lines that have **never been
compiled or run**, and it would fail if it were.

It is absent from `tests/meson.build`, and it is the only file in the repository that includes
`<gtest/gtest.h>` — this project does not use gtest anywhere, and gtest is not a build dependency.

Worse than unused, it encodes an expectation that **contradicts the implementation**. Its
`HandleInvalidFile` case asserts

```cpp
EXPECT_FALSE(RDeb822Source::ParseDeb822File(testFile, entries));
```

for a stanza missing `Suites` — but `ParseDeb822File` deliberately *skips* an incomplete stanza and
returns `true`. So wiring it up would produce a failure, not coverage.

Everything it covered is covered by tests that actually run, and the contract it got wrong is pinned
correctly by one of them:

| gtest case | covered instead by |
|---|---|
| parse / multiple entries | `test_deb822_separator`, `test_deb822_sourceparts` |
| write-and-read-back | `test_deb822_roundtrip` |
| comments | `test_deb822_roundtrip` (3 cases, incl. one with a colon) |
| extra fields | `test_deb822_roundtrip` (`Architectures`/`Languages`/`Targets`) |
| types round-trip | `test_deb822_types_roundtrip` |
| dialog population | `test_deb822_dialog` |
| **missing required field** | `test_deb822_separator:89` — asserts the read returns **true** and yields **0 records**, which is what the code actually does |

Full suite unchanged: **8/8 OK**, 0 compile errors, 0 undefined references.

---

Net effect of this and the previous cleanup: the PR went from **22 files / +2715** to
**15 files / +1378** — roughly half the diff — with identical functionality and the same 8 passing
tests. Every line removed was either never built or asserted something untrue.

What remains is the feature itself plus six tests that run: `common/rsource_deb822.{cc,h}`, the
`rsources`/`rgrepositorywin` changes, one line in `common/meson.build`, one in `po/POTFILES.in`, and
`tests/`.

tzh476:
One more in `3c5f0d83`, found by reviewing my own diff as you would: two of the six deb822 tests
hardcoded `/tmp` while the other four already honour `TMPDIR`, and one of those two used `mkdtemp`'s
result without a null check.

The null check matters more than it looks. With a genuinely unwritable `/tmp` (a read-only tmpfs
mounted over it — `chmod` is not enough, since tests run as root) and `TMPDIR` pointing elsewhere:

| | before | after |
|---|---|---|
| `/tmp` writable, `TMPDIR` unset | OK / OK | OK / OK |
| `/tmp` read-only, `TMPDIR=/alt` | **exit 1** / **exit 134 SIGABRT** | OK / OK |

`exit 134` is `test_deb822_types_roundtrip`: `mkdtemp` returns `nullptr` and `string(root)`
dereferences it. `test_deb822_sourceparts` exits 1 with a message instead, which is what both should
do — so the two problems are separable in the same experiment.

Worth admitting how I nearly missed it: my first version of that experiment was `chmod 000 /tmp`, and
**both the fixed and the unfixed tree passed it**, because root ignores the mode bits. A control the
broken version also passes proves nothing. I only caught it because I ran the unfixed arm instead of
trusting a green run on the fixed one.

Full suite: **8/8 OK.**

---

That is the third pass over my own diff, and the running total is:

| pass | effect | PR size after |
|---|---|---|
| six unbuilt, uncalled source files (incl. `rpackagemanager.h`, which you had deleted) | −1013 | 22 files / +2715 → 16 / +1702 |
| never-built gtest file that also asserted something untrue | −324 | 15 / +1378 |
| `TMPDIR` + `mkdtemp` null check | +16 | 15 / +1394 |

**Roughly half the original diff, no functionality removed, same 8 passing tests.** What is left is
`common/rsource_deb822.{cc,h}`, the `rsources`/`rgrepositorywin` changes, one line each in
`common/meson.build` and `po/POTFILES.in`, and six tests that actually run.

No rush on your side — @AsciiWolf said he is short on time for a month or two, and the GUI check is
his. I would rather the diff be small and honest whenever you do get to it.

mvo5:
Thanks for working on this! And sorry that it took me a little bit to look. Ideally we would cleanup the commit history a bit, 55 commits for this feature feel a bit much. But that is something to worry about later.

One small issue I noticed: I have a foo.source that had a signed-by: and then the inline pgp block. After disabling it and clicking safe this went away (the gpg block). It also set the source to "Enabled: yes" even though in the GUI its shown as not-enabled. And when I am not with the cursor on the row that has this repo its shown as an empty line. But when I highlight it it shows the correct content (except that it also shows the source as disabled which contradicts the "Enabled: yes" in the foo.source file itself.

mvo5:
So looking at this a bit more I wonder if we should simply split it as a starting point. I.e. split out the read side of the code and then mark any deb822 source read-only in the GUI. This should be a smaller diff and safer. Then we can do a followup with the write-path. We should look into switching to use libapts tagFile, iirc comments are a problem here but also iirc that was discussed a while ago and maybe we need to patch libapt (and we could vendor our own version of the pkgTagSection class if we have to until the fixed libapt is widely available).

mvo5:
Fwiw, I open #275 with a subset of this and #150 so that we can land it in smaller and more mangable chunks. Many thanks for your work on this, its greatly appreicated. Reviewing #275 and testing with various source files would be great too. Its read-only for now, the next steps would be to make sure the SourceRecord can represent the deb822 content byte-by-byte (the writing part is the thing that makes me most nervous :)

PR #275, opened 2026-09-05
many: initial read-only deb822 sources support

mvo5:
This PR adds some tweaks/work for the deb822 sources reading work in #274 and #150 and start with read-only support.

---

rsources: use libapt GetListOfFilesInDir instead of reimplmenting

Lets use apts GetListOfFilesInDir() instead of our own version
of it. Also add the beginging of tests around this which we will
need for the deb822 reading (gtest based).

Thanks to tzh476 and aybanda, who inspired this via PR#274 and
PR#150.

---

rsources: add a read-only version of deb822 parsing

To read apt's deb822 source records we now use libapt's pkgTagFile
and parse into our own SourceRecords. However we start with read
only support for now to keep the diff smaller.

Thanks to tzh476 and aybanda for the work on this in PR #274 and
PR #150. The test fixtures are taken from tzh476's tests in PR #274.

Co-authored-by: Stephen Smith <tzh476@users.noreply.github.com>

---

gtk: fix grayed out list view entries

---

gtk: add hint in rgrepositorywin.cc that deb822 is not
 editable

To avoid confusion this adds a hint in the repostiory dialog that
deb822 files are not editable yet. It also shows for one-line
sources what file they belong to.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,55 @@
+# Hacking on synaptic
+
+## Dependencies
+
+On Debian or Ubuntu the quickest way to get everything is
+
+    sudo apt build-dep synaptic
+
+The authoritative list is `Build-Depends` in `debian/control`.
+
+## Building
+
+    meson setup build
+    ninja -C build
+
+The binary is `build/gtk/synaptic`. Build options live in
+`meson_options.txt`; list them and their current values with
+
+    meson configure build
+
+and change one with e.g. `meson configure build -Dpkg_hold=true`.
+
+## Running the tests
+
+    ninja -C build test
+
+is the `make test` equivalent. For more control use meson directly:
+
+    meson test -C build --print-errorlogs
+    meson test -C build test_rsources
+    meson test -C build test_rsources --gtest-args='--gtest_shuffle --gtest_repeat=5'
+
+`test_rsources` uses GoogleTest and is only built when `gtest` is found.
+`-Dtests=enabled` makes a missing `libgtest-dev` a configure error, which is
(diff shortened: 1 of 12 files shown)


tzh476:
One review note on the deb822 URI path.

Skipping `SetURI()` for multi-value URIs is right, but `SetURI()` does three things and only one is multi-value-specific:

```cpp
S = SubstVar(S, "$(ARCH)", _config->Find("APT::Architecture"));
S = SubstVar(S, "$(VERSION)", _config->Find("APT::DistroVersion"));
if (URI[URI.size() - 1] != '/') URI += '/';
```

`sources.list(5)` documents `$(ARCH)` inside the deb822 section, so a user can write it in a `.sources` file. Since `rgrepositorywin.cc:473` displays `URI` directly, the same repository shows the expanded value from a `.list` and the literal `$(ARCH)` from a `.sources`.

Suggested, keeping the multi-value semantics:

```cpp
auto parts = FindMultiValue(Sec, "URIs");
for (auto &p : parts) {
   p = SubstVar(p, "$(ARCH)", _config->Find("APT::Architecture"));
   p = SubstVar(p, "$(VERSION)", _config->Find("APT::DistroVersion"));
}
rec.URI = APT::String::Join(parts, " ");
```

I left trailing-slash normalisation out: apt gives "suite must end with a slash" its own exact-path meaning, and I am not sure adding it here is correct.

Caveat on how I checked this: I reproduced both code paths in Python from the C++. I could not build or run the test suite (libapt is Debian-only, this is macOS), so I have not seen synaptic read a `.sources` file.

